Issue #87, opened 2013-01-28
goto definition changes colorcolumn


mbrochh:
I have this in my .vimrc:

```
set colorcolumn=80
highlight ColorColumn ctermbg=233
```

This renders a dark bar at column 80. When I use jedi-vim's goto definition, in all my tabs the colorcolumn suddenly is red, which I find very distracting.

Is this somehow my fault?

Issue #65, opened 2013-01-06
Jedi rename turns syntax highlighting off


Tinche:
After performing a jedi-rename (leader-r), the syntax highlighting turns off. I have to turn it back on manually.

syntonym:
In [`new_buffer()`](https://github.com/davidhalter/jedi-vim/blob/master/jedi_vim.py#L530-L534) the comment  ("sometimes syntax is being disabled and the filetype not set.") sounds just like this bug. Although there is no syntax highlighting, `g:syntax_on` is set to true. This sounds like a bug in vim? In [`_tabnew()`](https://github.com/davidhalter/jedi-vim/blob/master/jedi_vim.py#L539-L569) there is logic to switch to an existing tab instead of creating a new one. In [`new_buffer()`] there is no such logic, everytime `edit` is used. While I don't think `edit` should have such an effect, maybe switching to an existing buffer (with `:buffer`) would prevent that bug?

davidhalter:
I have no idea. I'm pretty sure that it is a VIM bug. I have no idea where it is actually coming from and I'm too lazy to narrow it down.

blueyed:
There have been fixes in this regard, especially https://github.com/davidhalter/jedi-vim/commit/e43012afc835e6db750ebbf299e452e9edb60f92.

So, does the problem still occur for you?

syntonym:
I've not encountered the bug anymore, seems fixed for me.

blueyed:
Cool, closing!

PR #101, opened 2013-02-22
Improved opening file in existing or new tab.

ricobl:
The `tabnew` command always opens a new tab, it is easier to let vim check if the file is open and reuse the tab or open a new tab by using the `tab drop <path>` command.

Maybe this could be applied to opening files in buffers also, using the `drop <path>` instead of the `edit` command. I think this could avoid some autocommands to be called unnecessarily.

Reference:
http://vim.wikia.com/wiki/Edit_a_file_or_jump_to_it_if_already_open

Possibly related issues:
https://github.com/davidhalter/jedi-vim/issues/65
https://github.com/davidhalter/jedi-vim/issues/87

### DIFF
--- a/plugin/jedi_vim.py
+++ b/plugin/jedi_vim.py
@@ -304,8 +304,7 @@ def rename():
 def tabnew(path):
     "Open a file in a new tab or switch to an existing one"
     path = os.path.abspath(path)
-    has_gui = vim.eval('has("gui")') == '1'
-    if has_gui:
+    if vim.eval('has("gui")') == '1':
         vim.command('tab drop %s' % path)
         return
 


gennaro-tedesco:
I am observing the same behaviour as described above, namely that most `goto` commands open new tabs even if already present. Is this issue still open or is it supposed to be fixed/working?

davidhalter:
No this is still an issue for me as well. However I feel like this is not a jedi-vim issue. Essentially we use `:tab drop` to open new tabs with a module's path. I feel like there's a bug in VIM that makes this not work properly sometimes. So feel free to have a look at the code.

I guess this should probably be reported in VIM. I have observed this for a long time and I'm unfortunately not sure how to reproduce it.

PR #339, opened 2014-12-19
Automatically fills import statement and display autocomplete popup

aliev:
there is a short screencast example: https://dl.dropboxusercontent.com/u/5837324/autoimport.mov

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -88,7 +88,8 @@ Example Installation Command using Pathogen:
 On Arch Linux, you can also install jedi-vim from official repositories as `vim-jedi
 <https://www.archlinux.org/packages/community/any/vim-jedi/>`__. It is also available
 on `Debian (≥8) <https://packages.debian.org/vim-python-jedi>`__ and `Ubuntu (≥14.04)
-<http://packages.ubuntu.com/vim-python-jedi>`__ as vim-python-jedi.
+<http://packages.ubuntu.com/vim-python-jedi>`__ as vim-python-jedi. On Fedora Linux,
+it is available as `vim-jedi <https://apps.fedoraproject.org/packages/vim-jedi>`__.
 
 Note that the `python-mode <https://github.com/klen/python-mode>`_ VIM plugin seems
 to conflict with jedi-vim, therefore you should disable it before enabling


davidhalter:
I'm still not 100% sure if it makes sense to include this. What do you think @blueyed?

autoload/jedi.vim
```diff
@@ -222,6 +222,16 @@ function! jedi#force_py_version_switch()
 endfunction
 
 
+function! jedi#auto_put_import()
+    " Auto put import statement after from module.name<space> and complete
+    if search('\<from\s\+[A-Za-z0-9._]\+\s*\%#\s*$', 'bcn', line('.'))
```
davidhalter: Shouldn't we check for whitespace before `from`? wouldn't that make sense?

blueyed:
I like the general idea!

But it will map "<Space>" and therefore cause some overhead for _every_ space you type.
The pattern could be improved, e.g. by anchoring it with `'^\s\+`, and there should be probably a fast-path handling, which is cheaper.

And then it's just some kind of basic snippet.
E.g. with UltiSnips you have a "from" snippet already, which expands to `from ${1:module} import ${2:Stuff}`.
When using this snippet, it would be good if there was a `smap` like this:

    smap <c-@> <del><C-x><C-o><C-r>=jedi#complete_opened()<CR>

Then you can do the following (<C-j> being my UltiSnips trigger):

    from<C-j>os<C-j><C-space>

I think this `:smap` is useful to have anyway - I'll add a PR for this.

autoload/jedi.vim
blueyed: The `\<` is a word boundary, but not necessarily whitespace, yes.

I would also anchor it at the beginning: `^\s\+

davidhalter:
Shouldn't we create an option that's more generic? Instead of `let g:jedi#auto_put_import = 1` we could use `jedi#auto_smart_completion` (which might be confusing, but something like it). Because basically you could add such checks for other things as well. (I don't have any good examples right now). Ideas?

blueyed:
:+1: for another name.
I suggest `jedi#smart_auto_mappings`.

aliev:
@davidhalter good idea!

davidhalter:
:+1:

blueyed:
@aliev 
Will you update the PR?
What do you think about having a fast-as-possible code path - at least by anchoring it to the beginning of the line?
I think the default for this should be 0 for now, while it's experimental.

davidhalter:
I actually think the default should be 1. It's not a very complicated thing and should just work. If it doesn't we shouldn't even merge.

aliev:
@blueyed @davidhalter oh sorry for big delay with my answer. I'm beginner in writing of vim of scripts. whether it is necessary to write some tests?

davidhalter:
So what I would like is:

1. Squash the commits (there are 3 `added doc`).
2. Use `jedi#smart_auto_mappings` as a name instead of `jedi#auto_put_import`.
3. Set the default to 1
4. A test would be nice. But you don't have to.

PR #347, opened 2015-01-05
Don't clear command line call signatures in completions()

wilywampa:
NeoComplete calls Jedi's `completefunc` frequently which clears Jedi's call signatures. This commit prevents command line call signatures from being cleared but doesn't affect the regular call signatures because they can errantly change the results of Jedi's completion function if not cleared.


davidhalter:
Could you add the description above into the code as a comment (to make clear why we added the check!).

wilywampa:
Done.

davidhalter:
Thanks!

PR #349, opened 2015-01-10
Remove duplicate signatures when show_call_signatures called

TKNGUE:
Jedi-vim showed duplicated signatures like below when show_call_signatures  is enable.
So I solve this probelm by remove duplicate signatures.

screenshot_from_2015-01-10 20 07 12

For reprodue:

``` python
improt numpy as np
np.abs() #cursor is in  parenthes
```

### DIFF
--- a/jedi_vim.py
+++ b/jedi_vim.py
@@ -265,6 +265,7 @@ def show_call_signatures(signatures=()):
 
     if not signatures:
         return
+    signatures = list(set(signatures))
 
     if vim_eval("g:jedi#show_call_signatures") == '2':
         return cmdline_call_signatures(signatures)


davidhalter:
If that is so easy to solve I have to check first if it's not an issue with Jedi itself.

TKNGUE:
OK, sir.
There is no plroblem so far in my environment.
In my confirmation that I run a vim with only jedi-vim on console,  This issue is observed.
I think this why abs is imported same file as names in different files(I don't know how jedi-vim works in detaild, so I don't have solid evidence)

joelmo:
I tried to edit this line to only return one element from the list, this works, but I don't know if there are any side-effects, there is probably a reason a list of call signatures is returned? https://github.com/davidhalter/jedi/blob/master/jedi/api/__init__.py#L587

blueyed:
@davidhalter
Any update from your side?

blueyed:
The example works for me:

    import numpy as np
    np.abs(

davidhalter:
Well, I just know that Jedi sometimes rightfully returns multiple functions. I would have to investigate why this happens here.

PR #360, opened 2015-02-08
Escape newlines in call signatures with show_call_signatures=2

wilywampa:
Currently the command line call signature treats '\n' in a string as a literal newline (e.g. in the call signature for `print()`) and this just escapes the newline so it is printed as literal '\n'.

### DIFF
--- a/jedi_vim.py
+++ b/jedi_vim.py
@@ -334,7 +334,7 @@ def cmdline_call_signatures(signatures):
         params = ['(' + ', '.join(p) + ')' for p in params]
     else:
         params = get_params(signatures[0])
-    text = ', '.join(params).replace('"', '\\"')
+    text = ', '.join(params).replace('"', '\\"').replace(r'\n', r'\\n')
 
     # Allow 12 characters for ruler/showcmd - setting noruler/noshowcmd
     # here causes incorrect undo history


davidhalter:
Hmm I agree with the issue, I disagree with the solution! :-) I think this should be part of Jedi, not jedi-vim.

davidhalter:
Examples? I don't find any.

wilywampa:
print():

    print(value, ..., sep = ' ', end = '\n', file = sys.stdout)

davidhalter:
Are there any other ones? I don't have VIM compiled with python 3, so I cannot see this one.

wilywampa:
I get the same call signature for `print` at work where there isn't a trace of python3, but this shows the issue:

```python
def foo(a, b="test\n"):
    pass

foo(
```

davidhalter:
@wilywampa I cannot reproduce this issue. At all.

Could it be that this got silently fixed in the last few jedi-vim/Jedi commits? Could you try to upgrade both?

davidhalter:
ping.

wilywampa:
Here's a demo using 054285e.

newlines

The contents of `.vimrc` for that demo:

```vim
set nocompatible noshowmode
set hidden hlsearch incsearch
set autoindent smartindent
set expandtab shiftwidth=4 softtabstop=4
set shortmess+=c
set undofile
set runtimepath+=$HOME/Downloads/dev/jedi-vim
set ttimeout ttimeoutlen=50
syntax enable
filetype plugin indent on
let g:jedi#show_call_signatures = 2
```

blueyed:
I see. It's an issue with `g:jedi#show_call_signatures = 2` (which I don't use).
I'll get this merged! Thanks!

btw: with `g:jedi#show_call_signatures = 2` it shows up as follows (independent of the newline):

    =`=jedi=0, =`=   (*_*a*_*, b="test\n") =`=jedi=`=

Why is that?
Looks like Jedi relies on some setting for this?

davidhalter:
Hmm. It shouldn't show up like that anywhere. The idea is that it shows up in the status line instead of the source code (which makes it obviously less of a hack).

wilywampa:
Mode 2 call signatures are meant to show up in the command line using Vim's `echon` command, not the status line. In the GIF I posted, I first show that my working copy has the change in this pull request applied and a call signature containing `\n` works (i.e. shows up in the command line). Then I checkout a clean copy and the call signature no longer works.

wilywampa:
@blueyed I think that's happening because I added `./jedi-vim` to my `rtp` but not `./jedi-vim/after`. Adding:
```vim
set runtimepath+=$HOME/Downloads/dev/jedi-vim/after
```
to `.vimrc` fixes that because it finds `./jedi-vim/after/syntax/python.vim` which sets up the conceal stuff.

blueyed:
Sorry, I've had an off-by-one error in my comment.. it showed up like this with `1`, and the missing `after` dir makes sense.

PR #363, opened 2015-02-15
Add 'winwidth' split option

petobens:
This allows to open a new split for "go to" in a vertical or horizontal split depending on the current window width.

### DIFF
--- a/jedi_vim.py
+++ b/jedi_vim.py
@@ -469,8 +469,12 @@ def new_buffer(path, options=''):
             'top': 'topleft split',
             'left': 'topleft vsplit',
             'right': 'botright vsplit',
-            'bottom': 'botright split'
+            'bottom': 'botright split',
+            'winwidth': 'vs'
         }
+        if user_split_option == 'winwidth' and vim.current.window.width <= 2 * int(vim_eval("&tw ? &tw : 80")):
+            for key, value in split_options.items():
+                split_options['winwidth'] = 'sp'
         if user_split_option not in split_options:
             print('g:jedi#use_splits_not_buffers value is not correct, valid options are: %s' % ','.join(split_options.keys()))
         else:
--- a/jedi_vim.py
+++ b/jedi_vim.py
@@ -472,9 +472,8 @@ def new_buffer(path, options=''):
             'bottom': 'botright split',
             'winwidth': 'vs'
         }
-        if user_split_option == 'winwidth' and vim.current.window.width <= 2 * int(vim_eval("&tw ? &tw : 80")):
-            for key, value in split_options.items():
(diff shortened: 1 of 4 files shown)


jedi_vim.py
```diff
@@ -469,8 +469,12 @@ def new_buffer(path, options=''):
             'top': 'topleft split',
             'left': 'topleft vsplit',
             'right': 'botright vsplit',
-            'bottom': 'botright split'
+            'bottom': 'botright split',
+            'winwidth': 'vs'
         }
+        if user_split_option == 'winwidth' and vim.current.window.width <= 2 * int(vim_eval("&tw ? &tw : 80")):
+            for key, value in split_options.items():
```
davidhalter: Why the loop?

petobens: Sorry about that. Place remove it if you intend to merge. Thank you!

davidhalter:
What is the visual difference of this versus the other ones? Also the option would need a further explanation in the README, but lets discuss it first.

Can anyone using splits in jedi-vim test this? And tell me a bit about it?

petobens:
@davidhalter  as it  can be seen in the GIF, the new split window is placed to the right or bottom depending on  the width of the window relative to 'textwidth'.

jump

davidhalter:
Where would it be placed with `bottom`? Under both splits?

If the window width is smaller than `textwidth` what happens? A normal split? True?

petobens:
>If the window width is smaller than textwidth what happens? A normal split? True? 

If it is smaller than the textwidth then a horizontal split opens:

tw

>Where would it be placed with bottom? Under both splits?

Yes, under both splits.

The last GIF shows what happens when one uses the `right` option: three vertical splits opens. If one uses `winwidth` option then we have what is seen in the GIF of my previous comment (after the first vertical split, there is a horizontal split).

right

davidhalter:
All right! Thanks for the explanation. I guess now we just have to explain it very well in the documentation. It should be clear that if the window is big enough a vertical split happens, if not a horizontal one.

Please also add the option (without description) to the README. And add yourself to `AUTHORS.txt`. Thank you!

jedi_vim.py
davidhalter: Fair enough, just add another commit. Please also rename `&tw` to `&textwidth`. I don't like the shortcuts in VIM script, they make everything way harder to read. (I know the `textwidth` option very well, but didn't know what it meant at first.)

petobens:
@davidhalter I think I have addressed your request. Note however that my english is not so good and my git skills even worse so please review the changes before merging. Thank you!

davidhalter:
Looks good! Thanks!

petobens:
Great.  Thank you

Issue #362, opened 2015-02-13
Rename variable with splits problem


petobens:
Consider the following minimal vimrc:

    set nocompatible

    let $ONEDRIVE_HOME = expand('C:/OD/Users/Pedro')
    let $DOTVIM = expand('$ONEDRIVE_HOME/vimfiles')

    let g:mapleader = ","

    set runtimepath+=$DOTVIM/bundle/jedi/

    if has('python3')
        let g:jedi#force_py_version = 3
    endif
    let g:jedi#use_tabs_not_buffers = 0
    let g:jedi#use_splits_not_buffers = 'right'

and, as in the GIF, open a `foo.py` file that contains

    x = 2
    print(x)

Now place the cursor on the `x` in the first line, type `,r` and once in insert mode type `y` and exit insert mode. You will se that jedi-vim correctly renames the variable but it also opens a split window. I expect the plugin not to open the split window.

![jedisplit](https://cloud.githubusercontent.com/assets/2583971/6196449/f3a41a84-b3b6-11e4-855b-735111262e70.gif)

petobens:
If I comment these [three lines](https://github.com/davidhalter/jedi-vim/blob/master/jedi_vim.py#L420-422) the problem disappears and the replacement seems to work fine. However I'm not sure if it has any side effects. Can you please check it?
Thank you.

davidhalter:
Hmm set shellslash seems to cause a lot of problems in many ways.

petobens:
@davidhalter indeed. Fortunately my issue was solved by 054285e2b7485d4adabfd4956806e6567eedebfe.

PR #364, opened 2015-02-20
Fix rename variable split issue #362

petobens:


jedi_vim.py
```diff
@@ -417,9 +416,6 @@ def rename():
                 vim.current.window.cursor = r.start_pos
                 vim_command('normal! cw%s' % replace)
 
-            result = new_buffer(window_path)
-            if not result:
-                return
             vim.current.window.cursor = cursor
```
davidhalter: Why do we not need these lines anymore?

petobens: @davidhalter  removing those lines fixes the problem when using the minimal vimrc in this [comment](https://github.com/davidhalter/jedi-vim/issues/362#issue-57663089). I don't know if it breaks something else (I've using jedi-vim with this change without any problem for the past few weeks).

davidhalter: Well the problem is, that I'm pretty sure you're breaking renaming over multiple files. So please include those lines again. Then we can talk about adding the other stuff.

petobens: @davidhalter  you are obviously right. Now I have no idea to fix my particular issue (renaming opens a new split of the same file) while allowing renaming over multiple files.  I'll try to find a fix but it might take some time.

davidhalter: Sure! This issue is known. If you find a fix, please let me know it. It must be something very strange and VIM specific.

petobens:
@davidhalter  sorry to bother you again. I wanted to create a new pull request with commit  https://github.com/petobens/jedi-vim/commit/c3ea309c60fd76522793cc9f8f3f2f4a36ea6d35 that allows to control the height of the documentation window. For some reason that commit is added to this pull request . Can you eventually merge that single commit instead of the whole pull request? Or do I need to somehow create a separate pull request? (Sorry for the question but I'm not good with git and I don't know how to do this). Thanks!

davidhalter:
@petobens It's not about being good at git. It's about knowing how github works :-)

Most people create a separate branch that they push to their own github repository where they also PR from. This is a good strategy, because you can do with master what you want and also update it, whereas updating the PR branches is not a good idea, because that would cause merge commits.
Github automatically adds new commits in a branch to current pull requests of that branch. This makes sense if you think about it: A PR is not something tied to a commit, but something that needs to evolve over time (criticism..). A PR on Github is basically just a branch that will or will not be merged.

However, if you want a clean repository (without merge commits), you somehow need to get back to an older state (`git reset COMMIT`) and then you can force push to overwrite this master branch (`git push -f`). If you want to create a new branch, use (`git branch new_branch_name`) and then push it `git push -u origin new_branch_name`. Hope I'm not mistaken. I tend to mix up git commands :-)

petobens:
@davidhalter  thank you for the detailed explanation. I'll try to create a separate pull request following you instructions.

blueyed:
@petobens 
Any update? (without having read into the issue)

petobens:
@blueyed  the commit that allows to change the height of the buffer showing the docs is fine and I believe can be merged. I need to somehow create a separate pull request for it (if you can do that for me that would be great, if not I''ll try to do it as soon as I get same time).

What I definitely cannot fix is the splitting issue when renaming variables. It would be really cool if you or someone else could try to fix it because I have no idea how to fix it without breaking renaming over multiple files.

blueyed:
@petobens 
I've extracted your commit and a change to it into https://github.com/davidhalter/jedi-vim/pull/402.

What I did was basically to create a new branch based on `dev` (`git checkout -b ... dev`), and then cherry-picked your commit (`git cherry-pick ...`).

jedi_vim.py
blueyed: @petobens 
Are you sure it breaks renaming in multiple files?

Multiple files should be handled by the `new_window` for `vim.current.buffer.name != r.module_path` above.

So from a quick glance it looks like these lines should be removed really.

I've started looking deeper into this - PR will follow.

petobens:
@blueyed thank you!! I didn't know about the `cherry-pick` command.

jedi_vim.py
petobens: @blueyed I pretty sure that when I tried it with multiple files after @davidhalter comment i found that it was, or at least it seemed, broken. However I cannot reproduce it now. I just test it with a very simple example and it works fine. I created a `foo.py` file with 
```python
def f():
    pass
```
and a `bar.py` file with `from foo import f`. When I rename the `f()` function in a the `foo` file then it is also renamed in the `bar` file.

Thank you for looking into this! :)

blueyed: @petobens 
Cool.
There are several places in this area that can fail - so you might have gotten confused.. :)
PR will follow soon.

blueyed:
I'm closing this in favor of the new PR #403.

Please check if that works for you, and fixes your initial issue.

petobens:
@blueyed your new PR fixes the issue partially because if I use `set shellslash` I need the change you mentioned in your last comment/note. Not really sure why but if I include it then it works.

blueyed:
@petobens 
What's the difference when you include it?
Then no new buffer / tab is opened? Or are there other issues then?

What is the output of `:python print(vim.current.buffer.name)` with and without this setting? How do the slashes look then?

And then the output from the following:

    python -c 'import jedi; print(jedi.Script("foo=1", 1, 1, "t.py").goto_assignments()[0].module_path)'

I think it might be good to wrap it with `noshellslash`, but therefore I'd like to understand what's the issue first.

btw: the default for this (when it's available) is `noshellslash` anyway. Are you setting this yourself, or through some plugin?

petobens:
@blueyed please see the gif below.

When I use `shellslash` several splits open when renaming (this does not happen if I use `noshellslash`). In the GIF you can also see the output of `:python3 print(vim.current.buffer.name)`.

Regarding `shellslash` I use it to interact properly with cygwin (maybe I should remove it and use `noshellslash`). However, wrapping around `noshellslash` seems to fix this particular issue.

blah

blueyed:
Interesting would be now what Jedi (probably) returns. See the snippet / code above.
And I'd appreciate the answer alone, without a gif to read it from.. :)

I guess it just uses backslashes.

Can you try this patch, please?

    diff --git i/jedi_vim.py w/jedi_vim.py
    index 21a2f3e..13ca392 100644
    --- i/jedi_vim.py
    +++ w/jedi_vim.py
    @@ -455,7 +455,7 @@ def rename():
                     if r.in_builtin_module():
                         continue
     
    -                if vim.current.buffer.name != r.module_path:
    +                if os.path.abspath(vim.current.buffer.name) != r.module_path:
                         result = new_buffer(r.module_path)
                         if not result:
                             return

petobens:
@blueyed  sorry for the gif, I thought that it would make the splitting issue clearer.

The code snippet returns `C:\OD\unix-tools\Vim\vim74\t.py`

Your patch works wonderfully and solves the problem. Thank you!! :)S

blueyed:
Thanks!

Committed in 054285e. Now this issue should be finally solved! :)

blueyed:
btw: I was aware of the splitting issue before - thanks for putting in the effort to create the gif!

davidhalter:
Thank you @blueyed and @petobens!

petobens:
Thank you @blueyed  and @davidhalter .

blueyed:
You're welcome! :octocat: 

Happy renaming! ;)

PR #366, opened 2015-03-03
Do not overwrite/re-define pythonString syntax regions

blueyed:
Use `containedin` instead with `jediFunction`.

This adopts any changes to the syntax definitions, instead of overwriting them.

I am not really sure how to test if it doesn't break anything.


davidhalter:
I will have to test it, but if it works -> good job!

blueyed:
I've rebased it against master: see #367.

PR #367, opened 2015-03-03
Do not overwrite/re-define pythonString syntax regions

blueyed:
Use `containedin` instead with `jediFunction`.

This adopts any changes to the syntax definitions, instead of overwriting them.

I am not really sure how to test if it doesn't break anything.

Followup to #366, this time against master - the dev branch appears to be outdated..


davidhalter:
Thanks so much. This looks really much better. So much clearer (and probably works with color schemes/syntax changes better).

I've also updated the dev branch. I generally prefer developing against it, but stopped doing it, because so many people send their PRs to master (they don't read `CONTRIBUTING.md`).

blueyed:
Thanks for merging!

You might want to change the default branch on Github maybe, but then it would get cloned by default, too.

PR #369, opened 2015-03-07
Set 'hide documentation' to jedi#documentation_command

pearofducks:
With default binding ("K") - `jedi#show_documentation()` behaves like a toggle. This behavior isn't replicated with a custom binding. This PR fixes that.


davidhalter:
Oh right! Thanks!

PR #372, opened 2015-03-14
Fix `E171: Missing :endif` error when jedi is not available

itchyny:
The command `echoerr` throws an error in Vim script, thus it causes the weird `Missing endif` error. This pull request fixes this issue.

### DIFF
--- a/initialize.py
+++ b/initialize.py
@@ -21,5 +21,5 @@ sys.path.insert(0, vim.eval('expand(s:script_path)'))
 try:
     import jedi_vim
 except ImportError:
-    vim.command('echoerr "Please install Jedi if you want to use jedi_vim."')
+    vim.command('echohl ErrorMsg | echomsg "Please install Jedi if you want to use jedi_vim." | echohl NONE')
 sys.path.pop(1)


itchyny:
@davidhalter, merge please!

davidhalter:
@itchyny Sorry, I was on holidays. I remember that we used to have a problem with a missing `endif`. I know that we fixed that somewhere along the line. Are you sure you're using the latest Jedi master branch?

I fail to see what you're fixing here.

itchyny:
Yes, I use the master branch and the problem persists.

itchyny:
Here's a proof.
screenshot 2015-03-22 21 57 37

davidhalter:
Hmm strange. I have to investigate. I still think that the missing `endif` comes from somewhere else and shouldn't be hidden by echoing the error in a different way.

itchyny:
Have you ever tested this plugin with a PC which you did not installed jedi?

itchyny:
How to reproduce the problem.
1. vim -u .vimrc
2. setf python

vimrc
```vim
set nocompatible
filetype off
set rtp+=~/.vim/path/to/jedi-vim
filetype plugin indent on
syntax enable
```

screenshot
jedi

davidhalter:
Well, I'm not saying you're wrong with your observation. Not at all. I just fear that this PR is just a workaround for an actual bug in jedi-vim, that should be fixed.

PR #373, opened 2015-03-14
Finish if compatible

itchyny:
A plugin must not change the global settings like this. Finish if it will not work.

### DIFF
--- a/plugin/jedi.vim
+++ b/plugin/jedi.vim
@@ -7,7 +7,7 @@
 
 " jedi-vim doesn't work in compatible mode (vim script syntax problems)
 if &compatible
-    set nocompatible
+    finish
 endif
 
 " jedi-vim really needs, otherwise jedi-vim cannot start.


davidhalter:
Well, but that wouldn't stop `autoload` and `ftplugin` from loading?

I agree that this is "hackish", but if you're not using `nocompatible`, you shouldn't be using jedi-vim. That's my take. In general I think that if you're using `compatible`, you should be using no plugin at all. :-) Are you really using it? I haven't met anyone, yet.

IMHO enabling `filetype plugin on` by default is the bigger change.

itchyny:
Definitely. I agree with you, those how sets `compatible` should not use plugins. I never set compatible. Many plugins does not work with `compatible` on therefore they should not be sourced. Any plugin should not change the `compatible` option unless it restore the value.

itchyny:
Also, `filetype plugin on` must be removed away.

davidhalter:
I disagree. :-) A plugin is allowed to use whatever it wants. Some plugins are just very annoying ;-)

With Jedi I try to find the middle ground. I don't want tell a hundred people that they need to use `filetype plugin on`, because that eats just a lot of time and is annoying for the people that are using Jedi as well. All of the people actually using plugins use those two options, so I don't mind activating them. It's just that VIM has a very annoying tradition of upholding backwards compatibility, which is finally broken with neovim.

I'm very annoyed with VIM's current scripting model, so I don't feel very guilty by including such commands.

I will not merge the current PR. If you want you can include both commands under `jedi#auto_initialization == 0`. That would actually make sense.

itchyny:
Even if you are annoyed by Vim's way, you are not allowed to do anything; that would be annoying for users or for other plugins. Changing the global environment should be avoided. They should tend to be modest. What if a plugin changes a global options randomly? What if a plugin sets nonumber since its author dislikes this option? A filetype specific plugin is allowed to change the local variables in the buffer of the specific filetypes. A statusline plugin is allowed to change the statusline option. A plugin manager plguin is allowed to change the runtimepath option. Do you still think that jedi is allowed to change the value of compatible option?

davidhalter:
I agree in general.

I disagree for those two options. By default Jedi also modifies other stuff that Jedi shouldn't modify (which I think is worse) with the `g:jedi#auto_vim_configuration` option. So I agree that both of those commands should be within an `jedi#auto_vim_configuration` check. But I will not disable that check, because it costs me so much in support time that I don't have (and I work already full time on Jedi)...

itchyny:
How often have you got issue reports regarding these before you included these commands? You seem to have troubled with much reports that you are tired of responding but I could not find that many issue reports in the bug tracker in this repository.

davidhalter:
Well, it's not directly related to those two commands, because they have been there since quite the start. 

My problem is rather that no one intentionally uses `compatible`, but some people tend to forget to include it. So I just fix it for them. I just don't care about that form of idealism, where no one really profits.

I have in the past repeatedly removed such options, where I had to admit that it could potentially destroy some different defaults. This is something that I will not remove. End of discussion. Sorry. ;-)

~ Dave

PS: I might keep it open though, to remember that I have to put `g:jedi#auto_vim_configuration` around it.

PR #383, opened 2015-04-08
add_goto_window: pass len of list and use min()

blueyed:
With less than `g:jedi#quickfix_window_height` entries in quickfix list,
this now only makes the window that large, effectively saving screen
space.

Maybe a new setting should be used for this instead, in case not having the same height always for the quickfix window might be an issue for someone?!


davidhalter:
I think that makes sense. Let me know if it doesn't work well for you.

PR #392, opened 2015-04-26
Fix popup-on-dot completion: use <C-p> to deselect the first entry

blueyed:
Without this, the <C-x><C-o> triggered by jedi-vim would cause the first
match to be inserted (and selected).  When using completopt=longest,
this was less so an issue.


blueyed:
I've rebased this, and submitted it against "dev": #396

PR #394, opened 2015-04-29
Use "echohl WarningMsg" instead of "echoerr" in no_jedi_warning

blueyed:
`echoerr` causes Vim to throw an error, but like the function name
indicates, this is only a warning.

### DIFF
--- a/jedi_vim.py
+++ b/jedi_vim.py
@@ -15,7 +15,7 @@ except ImportError:
 
 
 def no_jedi_warning():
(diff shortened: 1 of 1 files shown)


blueyed:
This PR is against master.. should I re-do it for dev, or can you cherry-pick / use it like this?

davidhalter:
No it's ok. I don't really care at the moment. I'm pushing everything to master as well.

davidhalter:
Thanks!

PR #397, opened 2015-04-29
Select mode mapping for completion trigger

blueyed:
This deletes the selection (by going to visual mode, and "c"), and then
triggers the completion.

This is useful with e.g. UltiSnips, when you want to start completion
with an expanded snippet argument selected.

Ref: https://github.com/davidhalter/jedi-vim/pull/339#issuecomment-97612632


davidhalter:
OK. I don't really understand the selection mode anyway, but it loks good!

Issue #374, opened 2015-03-18
completeopt always gets overwritten


wzyboy:
Hi David,

thanks for your great work!

I may find a little bug in jedi-vim. In ```README.rst``` you suggest that setting ```autocmd FileType python setlocal completeopt-=preview``` in ```~/.vimrc``` will disable the preview window. But this is not working for me.

After I added that line to my ```~/.vimrc```, I'm still getting annoying (for me, sorry) preview window when editing a python script. ```:verbose set completeopt``` returns:

      completeopt=menuone,longest,preview
        Last set from /usr/share/vim/vimfiles/ftplugin/python/jedi.vim

So my setting of ```completeopt``` in ```~/.vimrc``` is getting overwritten by ```jedi.vim```, and ```:scriptnames``` confirms my guess.

I find the option is getting set here: https://github.com/davidhalter/jedi-vim/blob/master/ftplugin/python/jedi.vim#L42-L43

So I added another line of ```let g:jedi#auto_vim_configuration = 0``` in ```~/.vimrc``` and ```completeopt``` does not get overwritten any more.

Is this a bug or is this by design? I guess this only occurs when people install jedi.vim in system directory.

davidhalter:
Hi Zhuoyun

This is by design. However, you're absolutely right, that changing the completeopt may not make sense and the documentation needs to be corrected.

If you want to modify it, you need to either `let g:jedi#auto_vim_configuration = 0` or modify it **after** Jedi has loaded. (Typically you invoke the plugins with pathogen, vundle, etc. After that you can just change the setting).

wzyboy:
> If you want to modify it, you need to either let g:jedi#auto_vim_configuration = 0 or modify it after Jedi has loaded.

The later approach is not available for me since I installed vim-jedi via system package manager and vimfiles in  ```/usr/share/vim/``` is always loaded at last.

I'll do with the former approach :-)

davidhalter:
I'm still going to reopen, because we should actually improve the documentation there.

blueyed:
It seems like autocommands for filetypes from ~/.vimrc (the `autocmd FileType python setlocal completeopt-=preview`) is being executed before filetype plugins.
You would have to put this into ~/.vim/after/ftplugin/python.vim (`setlocal completeopt-=preview`) - that's what the docs should then currently mention.

[...]

PR #398, opened 2015-04-30
auto-config: set completeopt only if it's the default

blueyed:
The setting is also moved from "ftplugin" to "plugin", which allows for
easier customization, e.g. via a FileType plugin.

Ref: https://github.com/davidhalter/jedi-vim/issues/374#issuecomment-97621368


davidhalter:
Makes perfect sense!

PR #401, opened 2015-05-01
Reorder jedi_vim.py / improved error handling

blueyed:


davidhalter:
Looks good. I haven't looked at all the details, but it should be fine! Thanks!

PR #403, opened 2015-05-02
Fix rename

### LINKED ISSUE #362: Rename variable with splits problem

blueyed:
This should fix #362 and other (non-reported) issues.

### DIFF
--- a/jedi_vim.py
+++ b/jedi_vim.py
@@ -425,16 +425,12 @@ def rename():
         vim_command('autocmd InsertLeave <buffer> call jedi#rename(1)')
         vim_command('augroup END')
 
-        vim_eval("let s:jedi_replace_orig = expand('<cword>')")
+        vim_command("let s:jedi_replace_orig = expand('<cword>')")
         vim_command('normal! diw')
         vim_command(':startinsert')
     else:
         # reset autocommand
         vim_command('autocmd! jedi_rename InsertLeave')
-
-        # Save original window / tab.
-        saved_tab = int(vim_eval('tabpagenr()'))
-        saved_win = int(vim_eval('winnr()'))
 
         replace = vim_eval("expand('<cword>')")
         vim_command('normal! u')  # undo new word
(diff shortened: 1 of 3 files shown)


jedi_vim.py
```diff
@@ -453,16 +457,29 @@ def rename():
...
+                        echo_highlight("Jedi-vim: failed to create buffer window for {}!".format(r.module_path))
+                        continue
+
+                buffers.add(vim.current.buffer.name)
+
+                # Save view.
+                saved_view = vim_eval('winsaveview()')
 
                 vim.current.window.cursor = r.start_pos
                 vim_command('normal! cw%s' % replace)
```
blueyed: @davidhalter 
I think that should be changed to `c%dl%s` % (len(replace), replace)` (or similar).
Otherwise it might fail / do something unexpected, depending on the `iskeyword` setting.
Is there something better to be used from Jedi's results?
This should also handle multibyte strings properly, but I think it would do.

blueyed:
I've refactored the renaming procedure a bit, and added a visual mode map for it.

davidhalter:
Awesome. I have tried to fix it for a long time :-)

What was the problem? I don't really understand that!

blueyed:
There were several issues.. :)

Starting from `new_buffer` being called too often, and not handling multiple windows in tabs properly etc.

davidhalter:
:+1:

PR #404, opened 2015-05-02
travis: add python 2.7 and 3.4 to test matrix

blueyed:


davidhalter:
I don't think that's necessary, since the tests are not written in Python anyway and it gives the impression that we're testing VIM with both Python 3 and Python 2.7 when actually the test runner just runs under both versions.

I don't care if you merge it, you can. But if you do get the tests working. :) If not you can close.

blueyed:
Yes, you're right.
The idea was to use different Python version for Vim, yes.
That would require to build Vim probably, but then allows to test a more recent version of Vim, too.
It has no priority for me - I'll leave it open for now.

davidhalter:
Auto close, because of the dev branch deletion. If you still want this, please open one again and reference this one.

PR #407, opened 2015-05-03
Better goto functionality

blueyed:
Based on discussions in https://github.com/davidhalter/jedi/issues/570.

Waiting for support in Jedi, but uses what's available currently for now.

### DIFF
--- a/autoload/jedi.vim
+++ b/autoload/jedi.vim
@@ -5,7 +5,6 @@ scriptencoding utf-8
 " ------------------------------------------------------------------------
 let s:deprecations = {
     \ 'get_definition_command':     'goto_definitions_command',
-    \ 'goto_command':               'goto_assignments_command',
     \ 'pydoc':                      'documentation_command',
     \ 'related_names_command':      'usages_command',
     \ 'autocompletion_command':     'completions_command',
@@ -17,9 +16,10 @@ let s:default_settings = {
     \ 'use_splits_not_buffers': 1,
     \ 'auto_initialization': 1,
     \ 'auto_vim_configuration': 1,
+    \ 'goto_command': "'<leader>d'",
     \ 'goto_assignments_command': "'<leader>g'",
+    \ 'goto_definitions_command': "''",
     \ 'completions_command': "'<C-Space>'",
-    \ 'goto_definitions_command': "'<leader>d'",
     \ 'call_signatures_command': "'<leader>n'",
     \ 'usages_command': "'<leader>n'",
     \ 'rename_command': "'<leader>r'",
@@ -183,16 +183,20 @@ endif
 " ------------------------------------------------------------------------
 " functions that call python code
 " ------------------------------------------------------------------------
+function! jedi#goto()
+    PythonJedi jedi_vim.goto(mode="goto")
+endfunction
+
 function! jedi#goto_assignments()
-    PythonJedi jedi_vim.goto()
+    PythonJedi jedi_vim.goto(mode="assignment")
 endfunction
 
 function! jedi#goto_definitions()
-    PythonJedi jedi_vim.goto(is_definition=True)
+    PythonJedi jedi_vim.goto(mode="definition")
 endfunction
 
 function! jedi#usages()
-    PythonJedi jedi_vim.goto(is_related_name=True)
+    PythonJedi jedi_vim.goto(mode="related_name")
 endfunction
 
 function! jedi#rename(...)
--- a/doc/jedi-vim.txt
+++ b/doc/jedi-vim.txt
@@ -22,12 +22,13 @@ Contents				*jedi-vim-contents*
 4. Usage				|jedi-vim-usage|
 5. Mappings				|jedi-vim-keybindings|
     5.1. Start completion		|g:jedi#completions_command|
-    5.2. Go to assignment		|g:jedi#goto_assignments_command|
-    5.3. Get original definition	|g:jedi#goto_definitions_command|
-    5.4. Show documentation		|g:jedi#documentation_command|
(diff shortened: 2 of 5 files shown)


doc/jedi-vim.txt
```diff
@@ -22,12 +22,13 @@ Contents				*jedi-vim-contents*
...
     5.1. Start completion		|g:jedi#completions_command|
-    5.2. Go to assignment		|g:jedi#goto_assignments_command|
-    5.3. Get original definition	|g:jedi#goto_definitions_command|
-    5.4. Show documentation		|g:jedi#documentation_command|
-    5.5. Rename variables		|g:jedi#rename_command|
-    5.6. Show name usages		|g:jedi#usages_command|
-    5.7. Open module by name 		|:Pyimport|
+    5.2. Go to definition (smart)	|g:jedi#goto_command|
+    5.3. Go to assignment		|g:jedi#goto_assignments_command|
+    5.4. Get original definition	|g:jedi#goto_definitions_command|
```
blueyed: Should say "Go to", too.

autoload/jedi.vim
```diff
@@ -183,16 +184,20 @@ endif
 " ------------------------------------------------------------------------
 " functions that call python code
 " ------------------------------------------------------------------------
+function! jedi#goto()
+    PythonJedi jedi_vim.goto(mode="auto")
```
davidhalter: Why `auto`? If you call it `goto_command`, why not call it `goto`?

davidhalter:
I think that's a good solution until it works in Jedi. 

Shouldn't we disable `goto_definitions` by default? I think that command is just confusing people and not really helping them.

blueyed:
> Shouldn't we disable goto_definitions by default?

What do you mean?
Remove it from the source completely? Or just not mention it in the docs anymore?

blueyed:
Is it OK to change the default mapping for "goto" to `<leader>j`?
Or should it stay at `d`?

davidhalter:
`j` is ok.

I don't think you need to deprecate it. I think it's good if you can still use it, but it's just something that shouldn't be enabled by default. I for one would actually still use it to debug Jedi.

sloria:
I'm confused: I've set `let g:jedi#goto_command = "<C-c>g"` in my vimrc, and when I open a Python file, I get the following deprecation warning:

```
'g:jedi#goto_command' is deprecated. Please use 'g:jedi#goto_assignments_command' instead. Sorry for the inconvenience.  
```

However, the documentation states:

```
NOTE: Deprecated. Use |g:jedi#goto_command| / |jedi#goto()| instead, which
currently uses this internally.
```

Also, the new `goto` still doesn't follow imports: is this expected?

blueyed:
Have you copy'n'pasted that?

The actual deprecation should be:

> "Jedi-vim: jedi#goto_definitions is deprecated, please use jedi#goto instead."

Do you have the maps defined manually?

sloria:
@blueyed Yes, I copy and pasted that. Seems to be coming from here: https://github.com/blueyed/jedi-vim/blob/bc21c5d179972fe2bf96312385ea99fc13522d42/autoload/jedi.vim#L41

sloria:
And I have remapped the go to command:

```
let g:jedi#auto_vim_configuration = 0
let g:jedi#goto_command = "<C-c>g"
```

sloria:
Looks like `goto_command` is listed in `s:deprecations`: https://github.com/blueyed/jedi-vim/blob/bc21c5d179972fe2bf96312385ea99fc13522d42/autoload/jedi.vim#L8

blueyed:
Oh, yeah.
I've totally missed that these are there.

And the code then maps it to "goto assignments", which explains the behaviour you're seeing.

And I've not seen it myself, because I manually create the mappings.

My added deprecation is this one: https://github.com/blueyed/jedi-vim/commit/331ce60ef07d7aeefab9f2bb1d758ae0582c0b96

@davidhalter 
So there was a `goto_command` before, which got deprecated?
It's probably been a while.. :)

autoload/jedi.vim
```diff
@@ -183,16 +183,27 @@ endif
...
+let s:displayed_deprecation = 0
 function! jedi#goto_definitions()
-    PythonJedi jedi_vim.goto(is_definition=True)
+    if !s:displayed_deprecation
+        echohl WarningMsg
+        echom "Jedi-vim: jedi#goto_definitions is deprecated, please use jedi#goto instead."
+        echohl None
+        let s:displayed_deprecation = 1
+    endif
+    PythonJedi jedi_vim.goto(mode="definition")
```
davidhalter: Can we not deprecate it? I think it's still useful for a few people (like me). I just don't want it to be there as a default, because 3 goto functions would definitely confuse people.

davidhalter:
> Also, the new goto still doesn't follow imports: is this expected?

In the current implementation, yes. We should just implement it properly in Jedi. Should be really easy actually.

autoload/jedi.vim
blueyed: Ok.

blueyed:
@davidhalter 
> `j` is ok.

I'll not change the default for now.

I've rebased and squashed the PR.
I'd like to get it merged shortly.

davidhalter:
Done. I think this is good for now. I haven't tested it personally, but I guess it should be fine.

blueyed:
Thanks!
I like your optimism! ;)

blueyed:
@sloria 
Following imports is possible in Jedi now.
See https://github.com/davidhalter/jedi-vim/issues/607 for the issue to integrate it in jedi-vim.

sloria:
Thanks for the heads up. That would be great functionality!

Issue #408, opened 2015-05-04
goto: use current window; respect / use 'switchbuf' setting in general


blueyed:
I find it irritating, when "goto" goes to another/existing window (in another tab), via `_newtab`.

I think that jedi-vim should use Vim's `switchbuf` setting in general:

    'switchbuf' 'swb'	string	(default "") global {not in Vi}

      This option controls the behavior when switching between buffers.
      Possible values (comma separated list):

         useopen	If included, jump to the first open window that
          contains the specified buffer (if there is one).
          Otherwise: Do not examine other windows.
          This setting is checked with |quickfix| commands, when
          jumping to errors (":cc", ":cn", "cp", etc.).  It is
          also used in all buffer related split commands, for
          example ":sbuffer", ":sbnext", or ":sbrewind".

         usetab	Like "useopen", but also consider windows in other tab
          pages.

         split	If included, split the current window before loading
          a buffer for a |quickfix| command that display errors.
          Otherwise: do not split, use current window.

         newtab	Like "split", but open a new tab page.  Overrules
          "split" when both are present.

Using `switchbuf` could eliminate some of the settings (`g:jedi#use_tabs_not_buffers` and `g:jedi#use_splits_not_buffers`), or rather could derive their defaults from it (`g:jedi#use_splits_not_buffers` allows for providing `left`, `right` etc.)

I find `jedi#use_tabs_not_buffers` (enabled by default) an odd setting by itself: e.g. for the quickfix list for the usages, there should be no extra handling, just use the standard quickfix mechanisms (which uses `switchbuf`).

Apart from that, `goto` should just use the window where it was called from, at least by default.

Some concrete questions:
 - should "rename" open new buffers? Probably yes, because these files have been changed.
 - `:Pyimport` should behave like "goto", i.e. not open a new buffer by default (according to `switchbuf`)?

davidhalter:
> should "rename" open new buffers? Probably yes, because these files have been changed.

Yes, I want to be able to undo the changes.

> I find jedi#use_tabs_not_buffers (enabled by default) an odd setting by itself: e.g. for the quickfix list for the usages, there should be no extra handling, just use the standard quickfix mechanisms (which uses switchbuf).

I'm not even sure where all of that stuff comes from. Most things have not been added by me, but by people that wanted a certain behavior. I therefore don't know how I feel about changing all of that.

[...]

PR #409, opened 2015-05-04
WIP: Improve buffer switching

blueyed:
Based on suggestions and discussion in https://github.com/davidhalter/jedi-vim/issues/408


blueyed:
Tests are failing, because the behavior changed.

Before adjusting them, I'd like to get feedback on this.

davidhalter:
Since this would be a change with big impact, I really need to play with it first. I don't have time for the next few days, so it might still take a while.

blueyed:
Sure.
Rebased it on master, and fixed a bug where it would not jump to the line in a new file.

davidhalter:
Since we have started the discussion in #408, I will continue it there.

PR #412, opened 2015-05-08
Change default for show_call_signatures to 2

blueyed:
This appears to be less error-prone.

E.g. with things like indenting plugins (https://github.com/hynek/vim-python-pep8-indent/pull/41).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -157,7 +157,7 @@ can improve the integrity of Vim's undo history.
 
 .. code-block:: vim
 
-    let g:jedi#show_call_signatures = "1"
+    let g:jedi#show_call_signatures = "2"
 
 Here are a few more defaults for actions, read the docs (``:help jedi-vim``) to
 get more information. If you set them to ``""``, they are not assigned.
--- a/doc/jedi-vim.txt
+++ b/doc/jedi-vim.txt
@@ -394,7 +394,7 @@ be disabled by setting this option to 0. Setting this option to 2 shows call
 signatures in the command line instead of a popup window.
 
 Options: 0, 1, or 2
-Default: 1 (Show call signatures window)
+Default: 2 (Show call signatures in command line)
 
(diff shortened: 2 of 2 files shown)


davidhalter:
Hmm true, I'm just not sure if it's what I want. I actually like the way of `show_call_signatures = 1` more, because it's what most IDE's do. I get however that it's a bit buggy. 

Since we don't have the same opinion about this, **I would like to gather community votes**. If a majority of people prefers `show_call_signatures = 2`, I would be willing to change.

oblitum:
+1 even thought less pretty.

blueyed:
:-1:, because I like them also more (having tried the one in the command line for a while now).

oblitum:
> :-1:, because I like them also more (having tried the one in the command line for a while now).

heh, there should be a meme for this situation.

davidhalter:
I'm closing this one now, doesn't seem to have generated a lot of interest.

nicolas33:
I vote `show_call_signatures = 3`, open a temporary window.

davidhalter:
Lol. Hell no. :)

nicolas33:
Why not? That's how docstring preview works. I guess that extending this feature for signatures is not a crazy idea.

davidhalter:
It would just add additional complexity. I already don't like specifying `show_call_signatures = 2`, because it's not a bool value. I believe in simple APIs, if possible.

The docstring preview already support call signatures. They are added there.

nicolas33:
Well, I don't find that `option = number` is complex.

Perhaps I took the issue the bad way. Since the signature is in the docstring preview, a good alternative could be to delay the closing of the popup until the closing parenthesis:

```
whatever. # <-- time the preview pops up.
whatever.call(with, any=parameter) # <-- time the preview gets closed. 
```

This makes sense because the docstring can still be usefull while typing the parameters.

PR #417, opened 2015-05-12
Fix rename behaviour with empty cword; honor b:changedtick

blueyed:
I think this is less fragile altogether.

It uses Vim's `b:changedtick` and skips the `:normal! A ' hack.

### DIFF
--- a/jedi_vim.py
+++ b/jedi_vim.py
@@ -73,7 +73,9 @@ def vim_eval(string):
 
 
 def no_jedi_warning():
-    vim.command('echohl WarningMsg | echom "Please install Jedi if you want to use jedi-vim." | echohl None')
+    vim.command('echohl WarningMsg'
+                '| echom "Please install Jedi if you want to use jedi-vim."'
+                '| echohl None')
 
 
 def echo_highlight(msg):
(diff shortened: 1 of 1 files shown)


jedi_vim.py
```diff
@@ -433,29 +435,39 @@ def get_params(s):
...
+            replace = vim_eval("expand('<cword>')")
+        else:
+            replace = None
+
         cursor = vim.current.window.cursor
-        vim_command('normal! u')  # undo the space at the end
+
+        # Undo new word, but only if something was changed, which is not the
+        # case when ending insert mode right away.
+        if vim_eval('b:changedtick != s:jedi_changedtick') == '1':
```
davidhalter: Hmm. I'm not sure if that actually works. Because you could just write "insert character, backspace" and you wouldn't have any real change, but the `b:changedtick` would have increased.

blueyed: It seems to work.  And the reason is that Vim considers this a change, too.

Try it with `:startinsert`, 'foo^H^H^H<Esc>' - `u` then will "undo" this.
But `:startinsert`, `<Esc>` won't be a separate undo step.

davidhalter: You are of course right. Thanks for correcting me. I thought this was about something different than what it is. Thanks!

PR #420, opened 2015-05-16
WIP: improve performance with call signatures

blueyed:
When using call signatures and moving around in insert mode, the performance is pretty bad.

I can see some potential in improving this.

This PR is work-in-progress and currently it just skips calling `cmdline_call_signatures`, if the signatures have not changed.

@davidhalter 
NOTE: it uses `str(...)` for comparison, because otherwise the signatures were not being considered to be equal.. a Jedi issue?

This first commit avoids the flickering in the cmdline, but still slows down moving around in insert mode.

Further ideas: we could only call out to Jedi in case the syntax name under the cursor has been changed.

### DIFF
--- a/jedi_vim.py
+++ b/jedi_vim.py
@@ -286,8 +286,13 @@ def show_documentation():
     return True
 
 
+last_displayed_signatures = None
+
 @catch_and_print_exceptions
 def clear_call_signatures():
+    global last_displayed_signatures
+    last_displayed_signatures = None
+
     # Check if using command line call signatures
     if vim_eval("g:jedi#show_call_signatures") == '2':
         vim_command('echo ""')
@@ -309,8 +314,6 @@ def clear_call_signatures():
             line = line[:match.start()] + match.group(2) + after
             vim.current.buffer[i] = line
     vim.current.window.cursor = cursor
-
-last_displayed_signatures = None
 
 @_check_jedi_availability(show_error=False)
 @catch_and_print_exceptions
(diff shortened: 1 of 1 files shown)


davidhalter:
Sorry, I don't really get the advantage of this? What does it make faster?

blueyed:
> What does it make faster?

For now it mostly only removes the flickering.

Is using `str()` for comparison (`str(signatures) == str(last_displayed_signatures)`) the current state of doing it?  Can Jedi be made to really compare them?

davidhalter:
I'm not sure. It just looks like a strange workaround to me. Let me analyze this later.

blueyed:
The idea is to skip generating / displaying the call signatures, if they would be the same.

blueyed:
Another idea: skip updating / rebuilding the call signature based on the Vim syntax (changes) and position of the current argument number.
I.e., as long as you are in "pythonString" there is no need to update the call signature.

davidhalter:
> Also as long as you are in (the same) "pythonString" there is no need to update the call signature at all. That should be detected in Vimscript.

But how do you know if you're in the same `pythonString`? I totally agree that we could improve performance there, but I'm not sure if that is really the problem. As I said previously: I think these ideas and changes should be within Jedi. There the caching would help all the implementations.

davidhalter:
I feel like you "fixed" this one yourself with a timer that you introduced. Performance improvements should however be part of Jedi IMO.

So I'm closing, because I feel like this is much better than back in the days.

Issue #421, opened 2015-05-16
Fix column detection with cmdline call signatures


blueyed:
Ref: https://github.com/davidhalter/jedi-vim/commit/cfd71a6b548ac26af8468071908f18f341467dc2

Is it necessary to display the signatures in the "correct" column?

I think the `:undojoin` in there is doing some unwanted things, and found the whole procedure really overkill.

If the correct column is useful then this should get fixed/simplified probably.

davidhalter:
cc @wilywampa

wilywampa:
I agree that `undojoin` causes some subtle but serious issues. In my fork of jedi-vim, I implemented a better `save_first_col` function which has the advantages of never switching windows, changing settings, or using `undojoin`. The disadvantage is that it only works with 3 or fewer windows. If 4 or more windows are open it just returns column 0. I think it's a good tradeoff for the number of side effects it eliminates.

It works just by looking and the window widths of the three (or fewer) windows doing some basic arithmetic to figure out what the layout is and where in the layout the current window is.

Here: https://github.com/davidhalter/jedi-vim/pull/422

wilywampa:
This can be closed since #422 was merged.

blueyed:
Thanks!

PR #422, opened 2015-05-17
Find first column using winwidth instead of wincmds/undojoin

wilywampa:
Change to address https://github.com/davidhalter/jedi-vim/issues/421. Use `winwidth` commands and geometry/math to find first column of current window instead of `wincmd`/`'eventignore'`/`undojoin` ugliness.

### DIFF
--- a/autoload/jedi.vim
+++ b/autoload/jedi.vim
@@ -341,11 +341,6 @@ endfunc
 
 function! jedi#configure_call_signatures()
     if g:jedi#show_call_signatures == 2  " Command line call signatures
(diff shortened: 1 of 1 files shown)


davidhalter:
Please check and merge @blueyed. Thanks! :) I'm on vacation for the next two weeks. Sorry, I'm a bit busy now.

blueyed:
I've tried it now finally, will merge it.  Thanks!

I've noticed some offset issues, but that appears to be caused by Jedi itself (e.g. https://github.com/davidhalter/jedi/issues/597).

PR #430, opened 2015-06-21
Use use_tabs_not_buffers=0 by default

blueyed:
Ref: https://github.com/davidhalter/jedi-vim/issues/408#issuecomment-101997691


davidhalter:
I agree!

PR #431, opened 2015-06-26
Fix ValueError with cmdline call signatures

blueyed:
Commit 8016665 (PR #360) escaped the newlines in `text`, which requires to use the
escaped version with `text.index` and `len`, too.

This fixes the following error:

    Traceback (most recent call last):
      File "…/jedi/jedi_vim.py", line 110, in wrapper
        return func(*args, **kwargs)
      File "…/jedi/jedi_vim.py", line 414, in cmdline_call_signatures
        left = text.index(params[index])
    ValueError: substring not found

/cc @wilywampa (author of PR #360)

### DIFF
--- a/jedi_vim.py
+++ b/jedi_vim.py
@@ -411,8 +411,9 @@ def cmdline_call_signatures(signatures):
 
     try:
         index = [s.index for s in signatures if isinstance(s.index, int)][0]
-        left = text.index(params[index])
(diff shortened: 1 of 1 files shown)


davidhalter:
Thanks @blueyed! Please test @wilywampa.

wilywampa:
This was definitely needed and works.

PR #436, opened 2015-07-01
Delayed call signatures

blueyed:
Add `show_call_signatures_delay` option

This will use CursorHoldI instead of CursorMovedI for displaying call
signatures, with a delay of 500ms by default.

This greatly enhances the experience with slow call signatures while typing
(depending on how fast Jedi is for the given context).

Ref: #420.

It just came to my mind that this addresses (my) main issues with the call signatures, which can be nasty when `Jedi` is slow on some more complex context while typing.

### DIFF
--- a/autoload/jedi.vim
+++ b/autoload/jedi.vim
@@ -26,6 +26,7 @@ let s:default_settings = {
     \ 'popup_on_dot': 1,
     \ 'documentation_command': "'K'",
     \ 'show_call_signatures': 1,
+    \ 'show_call_signatures_delay': 500,
     \ 'call_signature_escape': "'=`='",
     \ 'auto_close_doc': 1,
     \ 'max_doc_height': 30,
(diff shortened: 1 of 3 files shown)


davidhalter:
Hmm, that's actually a great idea. I will probably merge after I'm back from my holidays (in a week).

Is that normal? Would you also like to have that even if we had a better async solution?

blueyed:
> Is that normal? Would you also like to have that even if we had a better async solution?

Yes.
Jedi can be slow for complex constructs, and then it's a waste of resources in general to query it (uncached) again and again, while you are typing some string argument.
That's what https://github.com/davidhalter/jedi-vim/pull/420 is about to make better, and it would also include skipping the call to Jedi.

blueyed:
I've added some caching in Vim already (according to the cursor position between commas (arguments)): https://github.com/blueyed/jedi-vim/commit/89ccea711957a4a5f85194a858f4f7ca08623fb4

davidhalter:
I like the last commit a lot, however it just seems like a lot of work for such a small issue (displaying call signatures) :) 

I'm going to test this now for a bit and then I'm most probably going to merge.

davidhalter:
Hmm, it doesn't seem to work properly for me. I seem to get completions sometimes and then I don't and so on. The first one typically works. I couldn't work out where the issue was coming from.

davidhalter:
Going to be on holidays again for the next 3 weeks (with very limited Internet access. Don't expect me to answer quickly :)

blueyed:
Rebased on master and amended fixup commit.

Seems to work for me in general.  Feedback would be appreciated, e.g. if it still does not work.

davidhalter:
I like it now. I'm going to use it for a few days and let you know if it works fine. I tend to think it does.

Also feel free to ping me again when I forgot. There's unfortunately quite a lot of things going on here right now. (Also for other issues)

Thanks!

davidhalter:
Hmm I feel like this works quite well. Sometimes I don't get call signatures anymore. But I'm not sure if that's the fault of Jedi or jedi-vim. And then I'm not sure whether it's your patch or a general issue.

I guess I should just merge.

PR #441, opened 2015-07-21
s:init_python() fails on Windows

daisuzu:
Try to get the Python version from PATH on Windows,
fails to execute `system()` and exception occurs in vim.

Use double quotes to the shell argument, it works.


blueyed:
Thanks, I cannot verify this - so won't merge it (yet) myself.

Is it that `python -c 'import sys; sys.stdout.write(str(sys.version_info[0]))'` fails on Windows in general? Or only in Vim's `system()`?

daisuzu:
Thanks for response.
I think this is a Windows problem, the following error occurs when run directly in the cmd.exe.

* single quotes
single

* double quotes
double

blueyed:
Properly fixed via dc96886 now (#443).

PR #442, opened 2015-07-23
Add entry for Fedora Linux vim-jedi package

maxamillion:
Fedora packages jedi-vim as the package name vim-jedi, this just adds that information under the installation section of the README


davidhalter:
Thanks!

PR #443, opened 2015-07-25
Escape shell args properly on Windows

BohrShaw:
`cmd.exe` would strip off the first and last quote on a command.

### DIFF
--- a/autoload/jedi.vim
+++ b/autoload/jedi.vim
@@ -70,7 +70,7 @@ function! s:init_python()
         " avoids starting both of them.
 
         " Get default python version from interpreter in $PATH.
-        let s:def_py = system("python -c 'import sys; sys.stdout.write(str(sys.version_info[0]))'")
+        let s:def_py = system('python -c '.shellescape('import sys; sys.stdout.write(str(sys.version_info[0]))'))
         if v:shell_error != 0 || !len(s:def_py)
             if !exists("g:jedi#squelch_py_warning")
                 echohl WarningMsg


poppyschmo:
This fix worked for me on Windows with all `'shell'`-related options left at default (`'shellslash'`, `'shellxquote'`, etc.). Worked on posix/*sh too. Thanks.

blueyed:
See also https://github.com/davidhalter/jedi-vim/pull/441/files, which I prefer, since there's no `shellescape` required here.

Thanks for confirming the issue though!

blueyed:
Sorry, I've just noticed that using `shellescape()` would have been cleaner and more robust probably in general.

Please let me know in case some special `'shell'`-related option now causes issues with the change after 95b8dd7.

poppyschmo:
Tried a limited use-case scenario (MinGW), just to test, with `'shellslash'` on, `'shellxquote'` set to `"`, and `set shell=c:/\"Program\ Files\"/Git/usr/bin/bash.exe\ --login\ -i`. Worked before latest push. Also works using `shellescape()`.

blueyed:
@poppyschmo 
Thanks.

> Worked before latest push

The question however was rather if it still work, so _after_ the latest push..

poppyschmo:
Sorry, should have been clearer. No, the MinGW test case doesn't work after the latest push. Doubt many Windows users will be affected, though, since most use the default `%COMSPEC%`. Thanks.

blueyed:
I see.
Sorry for the mess and thanks for investigating.
Fixed in dc96886 now.

PR #453, opened 2015-08-17
Automatically fills import statement and display autocomplete popup

aliev:
Sorry for another one PR

### DIFF
--- a/doc/jedi-vim.txt
+++ b/doc/jedi-vim.txt
@@ -485,12 +485,12 @@ Default: "auto" (will use sys.version_info from "python" in your $PATH)
 ------------------------------------------------------------------------------
 6.12. `g:jedi#smart_auto_mappings`	            *g:jedi#smart_auto_mappings*
 
-When you start typing from module.name<space> jedi automatically
-fills import statement and display autocomplete popup.
+When you start typing `from module.name<space>` jedi-vim automatically
+adds the "import" statement and displays the autocomplete popup.
 
 This option can be disabled in the .vimrc:
 
-let g:jedi#smart_auto_mappings = 1
+`let g:jedi#smart_auto_mappings = 0`
 
 Options: 0 or 1
 Default: 1 (enabled by default)
--- a/doc/jedi-vim.txt
+++ b/doc/jedi-vim.txt
@@ -41,6 +41,7 @@ Contents				*jedi-vim-contents*
     6.9. completions_enabled		|g:jedi#completions_enabled|
     6.10. use_splits_not_buffers	|g:jedi#use_splits_not_buffers|
     6.11. force_py_version              |g:jedi#force_py_version|
+    6.12. smart_auto_mappings           |g:jedi#smart_auto_mappings|
 7. Testing				|jedi-vim-testing|
 8. Contributing				|jedi-vim-contributing|
 9. License				|jedi-vim-license|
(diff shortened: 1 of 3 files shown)


davidhalter:
I think that looks good. Agreed @blueyed ?

doc/jedi-vim.txt
```diff
@@ -481,6 +482,18 @@ Function: `jedi#force_py_version(py_version)`
 
 Options: 2 or 3
 Default: "auto" (will use sys.version_info from "python" in your $PATH)
+------------------------------------------------------------------------------
+6.12. `g:jedi#smart_auto_mappings`	            *g:jedi#smart_auto_mappings*
+
+When you start typing from module.name<space> jedi automatically
```
blueyed: s/jedi/jedi-vim/

And put the example into quotes (or backticks?), please.

doc/jedi-vim.txt
```diff
@@ -481,6 +482,18 @@ Function: `jedi#force_py_version(py_version)`
...
+fills import statement and display autocomplete popup.
```
blueyed: s/fills import/adds the "import"/
s/displays/displays the/

doc/jedi-vim.txt
```diff
@@ -481,6 +482,18 @@ Function: `jedi#force_py_version(py_version)`
...
+
+This option can be disabled in the .vimrc:
+
+let g:jedi#smart_auto_mappings = 1
```
blueyed: This should be 0.

blueyed:
Ok with me.

@aliev 
Can you fix the typos/doc and squash your commits?
You can then force-push the branch - there is no need to create a new PR.

Also, if you don't like to do this, I'd be happy to help you and do this when merging it.  Just let me/us know.

Thanks!

aliev:
@blueyed I updated the documentation.

> Also, if you don't like to do this, I'd be happy to help you and do this when merging it. Just let me/us know.

I think you can do it more faster than I. Thanks! :)

blueyed:
Ok.

What do you think about the following change to it?

It will limit the search to max. 50 bytes/chars (for the module name), and only matches directly behind it.

I found it odd that `from foo   _` would complete to `from foo    import _`.

We could normalize/strip the extraneous whitespace, but then again it might be a good workaround if you need to skip it, because it might be slow etc.

    index affc590..03bffd0 100644
    --- i/autoload/jedi.vim
    +++ w/autoload/jedi.vim
    @@ -425,7 +425,7 @@ endfunction
     
     function! jedi#smart_auto_mappings(is_popup_on_dot)
         " Auto put import statement after from module.name<space> and complete
    -    if search('\<from\s\+[A-Za-z0-9._]\+\s*\%#\s*$', 'bcn', line('.'))
    +    if search('\<from\s\+[A-Za-z0-9._]\{1,50}$', 'bcn', line('.'))
             " Enter character and start completion.
             return "\<space>import \<C-x>\<C-o>\<C-r>=jedi#complete_opened(".a:is_popup_on_dot.")\<CR>"
         endif

blueyed:
Edited the above comment.  Suggested new pattern:

    if search('\<from\s\+[A-Za-z0-9._]\{1,50}\%#\s*$', 'bcn', line('.'))

blueyed:
Also, `is_popup_on_dot` should be `0` always, shouldn't it?

blueyed:
I have tried adding a test:

    it 'smart import'
        exec 'normal ifrom os '
        Expect maparg(" ", "i") == 'jedi#smart_auto_mappings()'
        Expect getline('.') == 'from os import '
    end

But it fails:

    not ok 2 - completions smart import
    # Expected getline('.') == 'from os import '
    #       Actual value: "from os "
    #     Expected value: "from os import "=jedi#complete_opened(0)
    =jedi#complete_opened(0)

I have also tried using `feedkeys` for it.

blueyed:
Regarding the test, it appears that it's the `\%#` (cursor position) that does not get matched.  Doing the `normal` / `call feedkeys()` manually works.. :/

davidhalter:
Yeah testing VIM is a real pain :/

aliev:
@blueyed changed to new pattern in PR

blueyed:
I've managed to get it tested, which is possible when using a `<C-r>` mapping instead:
https://github.com/davidhalter/jedi-vim/compare/smart-import-mapping?expand=1

I'll merge/push this later when CI passes.

aliev:
@blueyed @davidhalter thank you guys! :)

aliev:
@blueyed oh sorry, I think I missed.It may be better to add <silent> for inoremap <buffer> <space> <C-R>=jedi#smart_auto_mappings()<CR>?

blueyed:
@aliev 
You're welcome!  Thank you, too! :octocat:

PR #454, opened 2015-08-19
Smart import mapping

aliev:


blueyed:
@aliev 
To quote you from #453:

> Sorry for another one PR

;)

There wasn't really a need for a new PR - but it's good maybe for looking at it in the future.

Anyway, thanks for adding and pushing this nice feature! :+1:

aliev:
@blueyed oh sorry, I think I missed.It may be better to add ```<silent>``` for ```inoremap <buffer> <space> <C-R>=jedi#smart_auto_mappings()<CR>```?

and it seems after changing to the ```<C-R>``` appears flashing cursor :-D

blueyed:
> It may be better to add <silent>

True.

> and it seems after changing to the <C-R> appears flashing cursor :-D

What do you mean?  Can you elaborate?

Also, I think we should use some delay: because if you keep typing the "import" yourself, it might get annoying quickly.
I think the mechanism from #436 could be used.  But then again an additional timeout will slow it down in general - so therefore you should probably get used to it, or disable it.. :)

aliev:
@blueyed alright :) what about silent? create another one PR?

blueyed:
> what about silent? create another one PR?

Committed/pushed already.. :)

I cannot reproduce the general flickering when pressing space (with or without `<silent>` and with both Vim and Neovim).
The `<silent>` fixed the display of the popup function (`jedi#complete_opened`?).
Does the `<silent>` fix the general flickering for you?

aliev:
> The <silent> fixed the display of the popup function (jedi#complete_opened?).

yes.

> Does the <silent> fix the general flickering for you?

nope :( cursor flickering occurs while pressing the space bar

aliev:
@blueyed I made a more detailed video: https://dl.dropboxusercontent.com/u/5837324/blink.mov

davidhalter:
Thanks so much @blueyed and @aliev. Really good work.

blueyed:
@aliev 
> it seems the problem occurs only when the cursor shape is enabled :)

Are you using Neovim?  If not, how do you configure it, and what version of Vim are you using?
I assume that it worked better with the `<expr>´ mapping?

I have different cursor shapes, too (natively supported by Neovim, custom config with Vim), and it works.  

What are you using to record the videos.  I've tried https://asciinema.org/, but the (i-beam) cursor shape is not handled/visible in the recorded video.

PR #455, opened 2015-08-19
Add jedi#reinit_python

blueyed:
This is meant to be used in case you want to re-init the Python version
used by jedi-vim, without restarting Vim.

I had this laying around.. it might be better to just export
`s:init_python` as `jedi#init_python`?!


davidhalter:
Hmm, I guess it's more concise than `jedi#init_python`. But I don't really care. Merge if you want!

PR #456, opened 2015-08-19
Add JediDebugInfo command: display used Python version

blueyed:
This command is meant to come in handy to debug problems with jedi-vim.
For now, it will display the Python version that jedi-vim is using.


davidhalter:
Very good idea. We could add `:scriptnames`, `:messages` and whatever if we want.

PR #457, opened 2015-08-19
nvim: add a warning in case the auto-detected version is not functional

blueyed:
This usually means that the `neovim` module is not installed, which
commonly happens in a virtualenv.


davidhalter:
Makes sense! I don't really have the means to test it, but it looks good.

PR #458, opened 2015-08-21
Add g:jedi#use_tag_stack feature for jedi#goto()

wilywampa:
I added a feature to use the tag stack instead of just the jump list. This is really convenient where you jump to a definition, then move around with all kinds of jumps to investigate whatever you jumped to, then can instantly jump back to the place `goto()` was called from with `<C-t>` instead of repeating `<C-o>` a bunch.

### DIFF
--- a/autoload/jedi.vim
+++ b/autoload/jedi.vim
@@ -34,7 +34,8 @@ let s:default_settings = {
     \ 'quickfix_window_height': 10,
     \ 'completions_enabled': 1,
     \ 'force_py_version': "'auto'",
-    \ 'smart_auto_mappings': 1
+    \ 'smart_auto_mappings': 1,
+    \ 'use_tag_stack': 1
 \ }
 
 for [s:key, s:val] in items(s:deprecations)
--- a/doc/jedi-vim.txt
+++ b/doc/jedi-vim.txt
@@ -511,7 +511,9 @@ Default: 1 (enabled by default)
 6.14. `g:jedi#use_tag_stack`			*g:jedi#use_tag_stack*
 
 Write results of |jedi#goto| to a temporary file and use the *:tjump* command
-to enable full *tagstack* functionality.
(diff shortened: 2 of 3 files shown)


blueyed:
Awesome idea!

davidhalter:
Nice! Why do we need a file for that? Is local memory not enough?

Also I would actually not implement the option unless someone specifically calls for it. We already have so many configuration options, if we don't really need it, we shouldn't add it.

wilywampa:
It would definitely be better to avoid using a temporary file, but as far I can tell there's no avoiding it. For example:

http://stackoverflow.com/questions/26291758/how-to-manually-push-stack-entry-into-vim-tagstack
http://vim.1045645.n5.nabble.com/Modifying-the-tag-stack-td1158229.html
http://vi.stackexchange.com/questions/2196/following-a-tag-in-a-vimscript

Every workaround involves writing to a temporary file.

As far as making it an option, my thinking is that read/write operations can be slow on some systems. I don't know how noticeable that would be on top of the actual work Jedi does so maybe that's not enough of an issue.

davidhalter:
Ok! Thanks for the explanation.

Read/write can be slow on systems, but Jedi itself does way more reads, typically.

davidhalter:
Is this finished?

wilywampa:
Yes! Had some trouble with the automated tests but I got there eventually. I can squash the commits if you'd like.

davidhalter:
So using <kbd>Ctrl</kbd>+<kbd>T</kbd> will get you back to the last one, right? Can you maybe add this to the documentation?

wilywampa:
I added an explanation of why it's useful and fixed the link to `|tagstack|` help.

blueyed:
Awesome!
Can you please rebase it on master, too?

davidhalter:
Nice! Thank you a lot.

Issue #460, opened 2015-09-08
Error for py files: `Please install Jedi if you want to use jedi-vim.`


robbyki:
Installed jedi-vim including Jedi using `git submodule update --init` from within repo after cloning with `clone --recursive` but still unable to get vim to detect jedi-vim whenever I try and open a python file.

blueyed:
You need to provide more information. E.g., which version of (Neo)Vim are you using?

robbyki:
Using vim 7.4 on Mac built from source with binary located in /use/local/bin compiled with python3 and with anaconda Python.

blueyed:
What is your output of `:ver`?
Does `:py3 print(1)` work?
Does `:py print(1)` work?
Try poking at the `import` handling (https://github.com/davidhalter/jedi-vim/blob/d634eb8a923f66b336f7af1a52a78bc8d77c0b6d/jedi_vim.py#L89) to see what the `ImportError` is about.

blueyed:
I've created #461, which might provide additional information for cases like this.

acganesh:
I had the same issue.  In my case (which may be different), the problem was that I had multiple installations of Python, and vim was detecting the wrong version.  

I was able to fix this on OS X 10.10 by recompiling macvim and explicitly declaring the `vi_cv_path_python` variable, which specifies the path to the correct Python binary to the build script.

```
vi_cv_path_python=/usr/local/bin/python \
./configure --with-features=huge  --enable-rubyinterp \
                  --enable-pythoninterp \
                  --enable-perlinterp \
                  --enable-cscope
```

Hope this helps someone.

blueyed:
@robbyki 
Closing this because of missing feedback.  It is likely a configuration issue on your side.

Please re-open or state that it should be re-opened, when you can provide more insight.  Thanks.

robbyki:
Sorry for the delay. Had the chance to pick this up again. Decided to try out a clean install vim latest vim. (see ver: below) and also testing out with latest python 3.5. Everything with vim compiled correctly and I installed jedi-vim using pathogen and I have it in my .vim/bundle directory. And I also successfully install jedi with pip. Still getting this message:

```
Please install Jedi if you want to use jedi-vim.
Error: jedi-vim failed to initialize Python: jedi#setup_py_version: Vim(py3file):Traceback (most recent call last): (in function jedi#init_python[3]..<SNR>63_init_python[50]..jedi#setup_py_version, line 18)```

[...]

PR #461, opened 2015-09-08
Pass ImportError to no_jedi_warning and display it there

blueyed:
Ref: https://github.com/davidhalter/jedi-vim/issues/460


davidhalter:
I think this is brilliant. Will help us a lot debugging certain things.

PR #464, opened 2015-09-27
make `key` and `val` script local

itchyny:
This pull request fixes a global variable environment pollution. You can confirm the leak by `:echo key` or `:echo val` after opening an Python file.


davidhalter:
Thanks mate!

Issue #469, opened 2015-10-05
Wrong Import Autocomplete In Comments


strin:
I was typing some text in comment that includes "from xxx". Once I press spacebar, the autocomplete fills in "import". 

<img width="592" alt="screen shot 2015-10-05 at 11 33 24 am" src="https://cloud.githubusercontent.com/assets/2183232/10289776/0fcb53d4-6b55-11e5-8af7-1806763bc5e7.png">

<img width="589" alt="screen shot 2015-10-05 at 11 33 33 am" src="https://cloud.githubusercontent.com/assets/2183232/10289778/12f014be-6b55-11e5-9e22-cbb94813a25a.png">

blueyed:
See also #465.  I'll leave it open because it's specific about comments/strings.

PR #470, opened 2015-10-05
smart_auto_mappings: look for `from` only at the start of a line

blueyed:
Fixes https://github.com/davidhalter/jedi-vim/issues/469.


davidhalter:
Thanks :) I guess you can just push such things, if you want.

Issue #474, opened 2015-10-13
Auto-increase cmdheight when with long sigs and show_call_signatures=2


fvisin:
When `jedi#show_call_signatures` is set to 2 a "Press ENTER or type command to continue" prompt appears every time a signature should be shown, unless `cmdheight` is set to 2. 

This workaround solves the problem by using a 2-lines command line, which is a waste of space and very ugly IMHO. I was wondering if it might be possible to automatically increase `cmdheight` before showing the signature, and decrease it back to the original value afterwards.

blueyed:
> a "Press ENTER or type command to continue" prompt appears every time a signature should be shown

I've never seen this.  Is it because the signature is too long?
Otherwise there seems to be a bug with your setup.
Please post your output of `:version`, and make sure to use the latest jedi-vim.

fvisin:
Yes, the problem arises when the signature is too long to be displayed. The best would be to temporary increase cmdheight of an amount that depends on the signature length. I am not an expert though, so I don't know if it is feasible.

davidhalter:
cc @wilywampa, because he originally implemented this.

wilywampa:
I'd also like to know your version and settings for `ruler`, `rulerformat`, `shortmess`, `showcmd`

Assuming you do have `ruler` set, this branch might be an improvement: https://github.com/wilywampa/jedi-vim/tree/callsig_columns

There are a couple complications that make:
- There is no way to check how much space vim is using in the command line for `showmode`, `ruler`, and `showcmd`
- Changing those settings in insert mode from Python corrupts the undo history so that's not acceptable

The code checks the length of the signature before displaying it and bails out if it thinks it will be too long (https://github.com/davidhalter/jedi-vim/blob/master/jedi_vim.py#L401).  The problem may be that I pulled the value of 12 columns that it uses to account for `ruler` and `showcmd` out of nowhere but I never have those settings enabled so I haven't run into it. I just actually measured how much space is used (assuming unset `rulerformat`) and it's 27 characters for both `showcmd` and `ruler` 17 characters for just `ruler`, and 10 for just `showcmd`.

I spent some time adjusting my tmux split size column-by-column with the different combinations of settings and found accounting for an additional 18 columns if `ruler` is set works pretty well, i.e. it refuses the draw the signature once the window gets even a column too narrow instead of giving the prompts like before. I made that change in the branch above.

blueyed:
> it refuses to draw the signature

[...]

PR #475, opened 2015-10-16
Truncate long command line call signatures

wilywampa:
Truncate command line call signature as described in https://github.com/davidhalter/jedi-vim/issues/474 when the full signature is too long to display.

### DIFF
--- a/jedi_vim.py
+++ b/jedi_vim.py
@@ -398,9 +398,10 @@ def cmdline_call_signatures(signatures):
         params = get_params(signatures[0])
     text = ', '.join(params).replace('"', '\\"').replace(r'\n', r'\\n')
 
-    # Allow 12 characters for ruler/showcmd - setting noruler/noshowcmd
-    # here causes incorrect undo history
-    max_msg_len = int(vim_eval('&columns')) - 12
+    # Allow 12 characters for showcmd plus 18 for ruler - setting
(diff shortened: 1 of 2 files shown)


blueyed:
Thanks!
Will give it a try!

A test case for this (even if it's just manual) would be nice.

davidhalter:
Nice! I don't see any issues.

@blueyed  Merge if you think it's finished.

blueyed:
Really nice.

@wilywampa 
1. I'd use a real ellipsis (`…`) instead of three dots.
That requires to mark the Python source file as utf8 (` -*- coding: utf-8 -*-`).
See https://github.com/blueyed/jedi-vim/commit/152a8dbaf0a8da7a0f963900a7bd72833fb2d264 - feel free to cherry-pick / include it.

2. It does not work right after `datetime.date(` anymore (which gets completed to `datetime.date()` for me via delimitMate.  Tested this on master, where it still works.

3. I've noticed some display issues with Neovim initially, when leaving insert mode and having resized the terminal window (changing the width).  I could not reproduce it now, and it might be a bug in Neovim after all.

blueyed:
FYI: 3. should be fixed with https://github.com/davidhalter/jedi-vim/pull/477.

wilywampa:
I merged in your ellipsis commits.

blueyed:
Cool, merged to master.

PR #477, opened 2015-10-18
Invalidate `show_call_signatures_delay` on InsertEnter

blueyed:
Followup to e241cf87fdccaa4f5eec48d4e4063920f2dfddea.
Ref: https://github.com/davidhalter/jedi-vim/pull/420.
Ref: https://github.com/davidhalter/jedi-vim/pull/475#issuecomment-149013542.

PR #479, opened 2015-10-18
Handle parenthesis with show_call_signatures' cache

blueyed:
After `o`, call signatures should show up after `abs(`.

Followup of https://github.com/davidhalter/jedi-vim/pull/477.

### DIFF
--- a/autoload/jedi.vim
+++ b/autoload/jedi.vim
@@ -376,15 +376,15 @@ function! jedi#show_call_signatures()
 
     " Caching.  On the same line only.
     if line == s:show_call_signatures_last[0]
-        " Check if the number of commas before or after the cursor has
-        " not changed: this means that the argument position was not
-        " changed and we can skip repainting.
+        " Check if the number of commas and parenthesis before or after the
+        " cursor has not changed since the last call, which means that the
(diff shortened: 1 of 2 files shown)


autoload/jedi.vim
```diff
@@ -403,7 +403,7 @@ function! jedi#configure_call_signatures()
         autocmd InsertEnter <buffer> let g:jedi#first_col = s:save_first_col()
     endif
     autocmd InsertEnter <buffer> let s:show_call_signatures_last = [0, 0, '']
-    autocmd InsertLeave <buffer> PythonJedi jedi_vim.clear_call_signatures()
```
davidhalter: F***. I actually thought this might be an issue, but at the same time I thought you might have changed it already.

blueyed: I had removed it intentionally, but now added it back - I think it's better to be explicit on resetting/invalidating the cache, although it gets done on `InsertEnter` now, too.

davidhalter:
Kudos for the test :)

blueyed:
> Kudos for the test :)

Thanks, appreciated! :)

Issue #478, opened 2015-10-18
jedi#goto_command gives traceback in python3


MiguelLatorre:
Reason:
python3 has no string.lowercase

Traceback (most recent call last):
  File "/home/mal/.vim/bundle/vim-jedi/jedi_vim.py", line 115, in wrapper
    return func(*args, **kwargs)
  File "/home/mal/.vim/bundle/vim-jedi/jedi_vim.py", line 258, in goto
    for _ in range(4)])
  File "/home/mal/.vim/bundle/vim-jedi/jedi_vim.py", line 258, in <listcomp>
    for _ in range(4)])
AttributeError: module 'string' has no attribute 'lowercase'

blueyed:
Thanks for the report.

It does not exist for Python 2 either.  A fix is available in https://github.com/davidhalter/jedi-vim/pull/480.

PR #480, opened 2015-10-18
Use string.ascii_lowercase instead of str.lowercase

blueyed:
Fixes https://github.com/davidhalter/jedi-vim/issues/478.

### DIFF
--- a/jedi_vim.py
+++ b/jedi_vim.py
@@ -254,7 +254,7 @@ def goto(mode="goto", no_output=False):
                         tagname = d.name
                         while vim_eval('taglist("^%s$")' % tagname) != []:
                             tagname = d.name + ' ' + ''.join(
-                                [random.choice(string.lowercase)
+                                [random.choice(string.ascii_lowercase)
                                  for _ in range(4)])
                         f.write('{0}\t{1}\t{2}'.format(tagname, d.module_path,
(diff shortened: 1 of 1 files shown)


jedi_vim.py
```diff
@@ -254,7 +254,7 @@ def goto(mode="goto", no_output=False):
                         tagname = d.name
                         while vim_eval('taglist("^%s$")' % tagname) != []:
                             tagname = d.name + ' ' + ''.join(
-                                [random.choice(string.lowercase)
+                                [random.choice(string.lower)
```
davidhalter: Why are we even joining the same thing four times? Besides: string.lower is a function and I don't even find a definition for `string.`

davidhalter: Ok. I just realized `string` is a module. `string.lowercase` doesn't exist in Python 3 anymore. It's named `ascii_lowercase` there.

You can just defined it manually:

```
>>> string.lowercase
'abcdefghijklmnopqrstuvwxyz'
```

blueyed: Yeah, the fix is not correct either.

`string` is imported as a module, and `string.lowercase` exists in Python 2: `'abcdefghijklmnopqrstuvwxyz'`.

`string.ascii_lowercase` should be used for both Python 2/3.

blueyed:
Pushed an amended commit.

PR #481, opened 2015-10-18
Improve use_tag_stack-hack: use context managers, short 'tags'

blueyed:
Use a NamedTemporaryFile context manager, which removes the temporary
file again, and use it as the only entry for the 'tags' setting to
improve lookup performance.

/cc @wilywampa

### DIFF
--- a/jedi_vim.py
+++ b/jedi_vim.py
@@ -8,9 +8,8 @@ import traceback  # for exception output
 import re
 import os
 import sys
-import string
-import random
 from shlex import split as shsplit
+from tempfile import NamedTemporaryFile
 try:
     from itertools import zip_longest
 except ImportError:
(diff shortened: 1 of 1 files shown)


wilywampa:
I actually used `NamedTemporaryFile` in my initial implementation: https://github.com/wilywampa/jedi-vim/commit/5041224c13f3f799db55c6d72137dbfde3a761f9

I changed to `tempfile()` because the Travis CI test didn't like `NamedTemporaryFile` but I didn't try it with a `prefix` argument, but that works so it's better.

If we're setting tags to just one file we can skip the while loop and addition of random characters to the tag name because there won't be any collisions by definition.

jedi_vim.py
```diff
@@ -262,10 +263,13 @@ def goto(mode="goto", no_output=False):
                                  for _ in range(4)])
                         f.write('{0}\t{1}\t{2}'.format(tagname, d.module_path,
                             'call cursor({0}, {1})'.format(d.line, d.column + 1)))
-                        f.seek(0)
```
davidhalter: @wilywampa @blueyed 
Do you guys know why this was there in the beginning? Why would you seek there?

blueyed: Maybe to trigger a flush?

blueyed:
> skip the while loop

Will add that.

jedi_vim.py
wilywampa: For this: http://stackoverflow.com/questions/1202848/python-tempfile-temporaryfile-cannot-be-read-why

Does `flush` do the same thing?

blueyed: `flush` is necessary for the file to be readable/filled by Vim.
The above SO answer might be only relevant for the same Python process maybe (a comment states that `flush` does not work in this case)?!

blueyed:
> If we're setting tags to just one file we can skip the while loop and addition of random characters to the tag name because there won't be any collisions by definition.

Done.

davidhalter:
Is this one ready? (There are merge conflicts btw.)

blueyed:
Rebased and amended it.
Should be good from my side.

@wilywampa ?

blueyed:
I'm going ahead and merge it, hopefully fixing #482.

Issue #482, opened 2015-10-19
Latest update breaks Neovim capability for goto function assignments & definitions?


perks:
Recently updated the plugin, and found this error being fired off when trying to take advantage of the ```<leader> g``` and ```<leader> d``` behaviour.

Python installed correctly (```echo has("python")```) as well as running ```:JediDebugInfo``` returns: ```Using Python version: 2```

The error mentions something about tags, I've never generated tags before for this project and it was working flawlessly before

Bellow is traceback: 

```
Traceback (most recent call last):
  File "/home/chris/.vim/plugged/jedi-vim/jedi_vim.py", line 119, in wrapper
    return func(*args, **kwargs)
  File "/home/chris/.vim/plugged/jedi-vim/jedi_vim.py", line 267, in goto
    vim.command('tjump %s' % tagname)
  File "/usr/local/lib/python2.7/dist-packages/neovim/api/nvim.py", line 119, in command
    return self._session.request('vim_command', string, async=async)
  File "/usr/local/lib/python2.7/dist-packages/neovim/api/common.py", line 230, in request
    return walk(self._in, self._session.request(name, *args, **kwargs),
  File "/usr/local/lib/python2.7/dist-packages/neovim/msgpack_rpc/session.py", line 90, in request
    raise self.error_wrapper(err)
NvimError: Vim(tjump):E433: No tags file
```

blueyed:
I'd guess that https://github.com/davidhalter/jedi-vim/pull/481 fixes it.

Can you try and confirm that?
Otherwise this is likely to be merged soonish anyway.

blueyed:
#481 is merged now, please re-try.

blueyed:
If it does not work, please provide a new traceback.

Thanks!

perks:
Pulled and retried, still suffering same error:

```
Traceback (most recent call last):
  File "/home/chris/.vim/plugged/jedi-vim/jedi_vim.py", line 118, in wrapper
    return func(*args, **kwargs)
  File "/home/chris/.vim/plugged/jedi-vim/jedi_vim.py", line 264, in goto
    vim.command('tjump %s' % tagname)
  File "/usr/local/lib/python2.7/dist-packages/neovim/api/nvim.py", line 119, in command
    return self._session.request('vim_command', string, async=async)
  File "/usr/local/lib/python2.7/dist-packages/neovim/api/common.py", line 230, in request
    return walk(self._in, self._session.request(name, *args, **kwargs),
[...]

PR #489, opened 2015-10-30
Fix handling of temporary tags file on Windows

wilywampa:
Using `NamedTemporaryFile` doesn't work in Windows because Windows can't read the file until the file handle is closed. Making a simple context manager to delete a temporary file after using the tag stack fixes https://github.com/davidhalter/jedi-vim/issues/482.

### DIFF
--- a/jedi_vim.py
+++ b/jedi_vim.py
@@ -9,7 +9,7 @@ import re
 import os
 import sys
 from shlex import split as shsplit
-from tempfile import NamedTemporaryFile
+from contextlib import contextmanager
 try:
     from itertools import zip_longest
 except ImportError:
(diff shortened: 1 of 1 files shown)


jedi_vim.py
```diff
@@ -205,6 +205,17 @@ def completions():
         vim.command('return ' + strout)
 
 
+@contextmanager
+def tempfile(content):
+    # Using this instead of the tempfile module because Windows won't read
+    # from a file not yet written to disk
```
blueyed: Would it be possible to `flush()` it as an alternative?

wilywampa: That's what we were doing before which didn't work on Windows, unless I'm misunderstanding.

blueyed:
Thanks!

PR #497, opened 2015-11-07
Fix signatures only working in a single buffer

wilywampa:
The autocommands that display call signatures are reset globally any time `ftplugin/python/jedi.vim` is sourced so call signatures only work in the most recently loaded buffer. This change only resets autocommands in the current buffer. All of the `jedi_call_signatures` autocommands use `<buffer>` so that's what we want.


davidhalter:
Thanks a lot!

PR #499, opened 2015-11-11
Fix parameters missing from command line call signatures

wilywampa:
Parameters (or ellipsis) should show up in call signatures even when the index of the current argument is unknown. This mostly happens when there are more arguments typed than in Jedi's idea of the signature. Currently it just shows empty parentheses even if there are parameters (this applies to command line signatures only).

PS:  When the testing fails in `test/signatures.vim` I get a `UnicodeEncodeError` from `conftest.py` ever since using the ellipsis character.

### DIFF
--- a/jedi_vim.py
+++ b/jedi_vim.py
@@ -450,10 +450,12 @@ def cmdline_call_signatures(signatures):
         max_msg_len -= 18
     max_msg_len -= len(signatures[0].call_name) + 2  # call name + parentheses
 
-    if max_msg_len < 0:
+    if max_msg_len < (1 if params else 0):
         return
     elif index is None:
-        pass
+        text = escape(', '.join(params))
+        if params and len(text) > max_msg_len:
+            text = ELLIPSIS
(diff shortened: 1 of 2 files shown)


davidhalter:
Nice. I think the idea is good. I'm not 100% sure about the usability, but since you are the one using this feature and not me, I think you should just try it out :)

@blueyed Maybe you can check this as well (and merge if you want).

blueyed:
Looks good!

@wilywampa 
> PS:  When the testing fails in `test/signatures.vim` I get a `UnicodeEncodeError` from `conftest.py` ever since using the ellipsis character.

Happens with Python 2 only: created a new issue to track it at https://github.com/davidhalter/jedi-vim/issues/500.

wilywampa:
This got closed as I was about to write this which is good. I was just going to clarify that this was just a bug fix. This is how the standard call signatures work and how the command line call signatures worked because I put in the ellipsis stuff. I just missed this case in the refactoring.

PR #505, opened 2015-11-26
Update jedi.vim

zky001:
for OS X system,change more easy way


blueyed:
This would trigger completions on normal `<Space>`?
I don't think that makes sense.

davidhalter:
Yeah, this is definitely not what we want.

PR #507, opened 2015-12-05
[RFC] split jedi into external process

nakamuray:
Currently, jedi.vim load jedi module using vim's builtin python interpreter, which restrict module explore path only from that python interpreter and virtualenvs created for that interpreter.

So, for example, if I use vim linking with python 2.7 and working on virtualenv created for python 3.4, jedi.vim doesn't list completion candidates installed on the venv.

To fix the problem, I propose splitting jedi loading process into external process and make jedi.vim communicate it through pipes.
Because any python interpreter can execute this external process, user can freely choose which python interpreter to use and get expected completion results.
It assume that "python interpreter user expected" is a command executed as a "python" for now.

I tested this code on linux and osx.
It may cause some problems on windows environment. (cause I have no windows installation, I couldn't test it.)

### DIFF
--- a/jedi_remote.py
+++ b/jedi_remote.py
@@ -26,15 +26,8 @@ class PoorRPC(object):
         for line in iter(self.input.readline, ''):
             data = json.loads(line)
 
-            func = getattr(self, 'func_' + data['func'])
-
-            if func is None:
-                self._write_output({
-                    'code': 'ng', 'message': 'no such func',
-                })
-                continue
-
             try:
+                func = getattr(self, 'func_' + data['func'])
                 ret = func(*data['args'], **data['kwargs'])
             except Exception as e:
                 result = {'code': 'ng', 'message': repr(e)}
--- a/jedi_remote.py
+++ b/jedi_remote.py
@@ -1,0 +1,99 @@
+import json
+import os
+import sys
+
+# add path for jedi
+sys.path.append(
+    os.path.join(os.path.dirname(os.path.abspath(__file__)), 'jedi'))
+
+import jedi
+
+
+class PoorRPC(object):
+    def __init__(self, input=None, output=None):
+        if input is None:
+            input = sys.stdin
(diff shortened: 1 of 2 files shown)


davidhalter:
Thanks so much mate! I'm not 100% sure yet what I think about it, but what do you think about https://github.com/davidhalter/jedi/issues/385? Which would generalize this for all kinds of Jedi usages?

nakamuray:
I think this discussion forgot the case that user may install and use same python version on multiple places, but yes, it generalized (and complicated) version of mine.
I don't know about cache, though.

davidhalter:
>  forgot the case that user may install and use same python version on multiple places

What do you mean by that?

> I don't know about cache, though.

Caching is an issue, but you wouldn't need to care about it, I could handle that. At the moment it's not handled anyway, so the server wouldn't even need to handle that in the beginning.

----------------------------------

Are you interested in creating a general async server for Jedi?

If so one of the biggest issues is actually that sometimes you may want to not call certain methods, because performance would be improved. What would you think about that? Keeping a session?

nakamuray:
> What do you mean by that?

For example, User, who has /usr/bin/python2.7, also install python 2.7 under his home directory, say, /home/nakamuray/python27/bin/python2.7.
In this case, just switching between python "versions" is not sufficient, as it couldn't distinguish /usr/bin/python2.7 and /home/nakamuray/python27/bin/python2.7.
I thinks server should receive and switch "python interpreter path", not version.

> Are you interested in creating a general async server for Jedi?

No. Async, multiple python version jedi server is beyond my needs.
But for that future, I created a new branch, which maintain jedi API compatibility and implement remote object method calling.
https://github.com/nakamuray/jedi-vim/tree/rpc2

tell-k:
+1

nakamuray:
open a new PR based on rpc2 branch. close it.

PR #511, opened 2015-12-15
split jedi into external process, v2

nakamuray:
It is second version of jedi RPC work. (v1, #507)

This time, RPC version of jedi object preserve original jedi compatible API.
I also added config option to toggle RPC code (disabled by default).

### DIFF
--- a/jedi_remote.py
+++ b/jedi_remote.py
@@ -1,0 +1,127 @@
+from __future__ import print_function
+
+import json
+import os
+
+import jedi.api
+
+from subprocess import Popen, PIPE
+
+
+class JediRemote(object):
+    '''Jedi remote process communication client
+
+    This class provide jedi compatible API.
+    '''
+    python = 'python'
+    remote_command = 'jedi-remote-command.py'
+
+    def __init__(self, python=None):
+        if python is not None:
+            self.python = python
+
+        self._process = None
+
+    def __del__(self):
+        if self._process is not None:
+            # XXX: why does this raise exception?
+            #self._process.terminate()
+            self._process = None
+
+    @property
+    def process(self):
+        if self._process is None or self._process.poll() is not None:
+            cmd = os.path.join(
+                os.path.dirname(
+                    os.path.abspath(__file__)), self.remote_command)
+            self._process = Popen([self.python, cmd], stdin=PIPE, stdout=PIPE)
+
+        return self._process
+
+    def remote_object_getattr(self, id, *args, **kwargs):
+        return self._call_remote('remote_object_getattr', id, *args, **kwargs)
+
+    def remote_object_setattr(self, id, *args, **kwargs):
+        return self._call_remote('remote_object_setattr', id, *args, **kwargs)
+
+    def remote_object_call(self, id, *args, **kwargs):
+        return self._call_remote('remote_object_call', id, *args, **kwargs)
+
+    def remote_object_repr(self, id, *args, **kwargs):
+        return self._call_remote('remote_object_repr', id, *args, **kwargs)
+
+    def free(self, id):
+        return self._call_remote('free', id)
+
+    def __getattr__(self, name):
+        return self._call_remote('get_from_jedi', name)
+
+    def _call_remote(self, func, *args, **kwargs):
+        output = json.dumps({'func': func, 'args': args, 'kwargs': kwargs})
+
+        self.process.stdin.write(output.encode('utf-8'))
+        self.process.stdin.write(b'\n')
+        self.process.stdin.flush()
+
+        input = json.loads(self.process.stdout.readline().decode('utf-8'),
+                           object_hook=self.remote_object_hook)
+
+        if input.get('code') == 'ok':
+            return input['return']
+
+        elif input.get('code') == 'ng':
+            exception_name = input['exception']
+            e_args = input['args']
+
+            # try to find from jedi first
+            exception_class = getattr(jedi.api, exception_name, None)
+            if exception_class is None:
+                # ... and then, try to find from builtin
+                import __builtin__
+                exception_class = getattr(__builtin__, exception_name, None)
+
+            if exception_class is None:
+                exception_class = Exception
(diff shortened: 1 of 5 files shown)


jedi_remote.py
```diff
@@ -0,0 +1,127 @@
...
+
+from subprocess import Popen, PIPE
+
+
+class JediRemote(object):
+    '''Jedi remote process communication client
+
+    This class provide jedi compatible API.
+    '''
+    python = 'python'
```
blueyed: Can this be made configurable in Vim?
May `'use_rpc'` could be changed to not use this RPC method when empty, but would otherwise be the path / command to use?

jedi_remote.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def __init__(self, python=None):
+        if python is not None:
+            self.python = python
+
+        self._process = None
+
+    def __del__(self):
+        if self._process is not None:
+            # XXX: why does this raise exception?
```
blueyed: What exception does it raise?

jedi_remote.py
```diff
@@ -0,0 +1,127 @@
...
+            #self._process.terminate()
```
blueyed: What exception does it raise?

blueyed:
Very interesting, thanks!

How does it compare with regard to performance?
Does it have any benefit over Neovim's method, which also uses an external Python process?

jedi_remote.py
nakamuray: Yes, possible.
I'll do that.

jedi_remote.py
nakamuray: `Exception AttributeError: "'NoneType' object has no attribute 'SIGTERM'" in <bound method JediRemote.__del__ of <jedi_remote.JediRemote object at 0x7f2f470ed250>> ignored`

nakamuray:
> How does it compare with regard to performance?

Apparently, it has some performance penalty.
I choose JSON as a serialization method because it is in the standard library. (and I doubt pickle could use to communicate between python 2 and 3.)
But it is not a fastest serialization format.

It may be possible to get some speed by including known-to-be-used variables to the RemoteObject, which reduce RPC round trip.

> Does it have any benefit over Neovim's method, which also uses an external Python process?

It works on vim!!!

...  In fact, there is not so much benefit over neovim's. If I had to say anything,

- neovim seems to switch python interpreter at `:python` (or `:python3`) level. So, if you have vim plugin which need python3.4, but want to use python3.5 to get completion candidates, you can't satisfy both need. 
  My code can switch python version independently.

- It may be possible to extend it to run jedi process for each buffers using multiple python versions.
  For example, you'll spawn python3.4 jedi process for buffer A, python3.5 jedi for buffer B and python3.5 with another virtualenv for buffer C to get appropriate completion result.

- Currently, it is not require any external packages to run. No need to install "neovim" package for each virtualenvs you work on.

davidhalter:
Thanks for all the work!!

I think this is quite an ok solution. However I don't see what it actually improves besides calling arbitrary python versions (which is great!).

What I would want for an async solution is an async server that is not living within jedi-vim. Such a server should be a part of Jedi! There's a good reason for it: This http server can be used for only within jedi-vim, while it could benefit a few other plugins as well.

Hence once we're talking about a more general solution, we would need to talk about session handling (because some parts of the API are lazy). How do we keep that state? How do we do garbage collection on such sessions... etc. etc.

In conclusion: I will not be merging this. It's just - not what we need in the long term. I might merge something that at least contains an async function that makes completion "faster". But I don't see big usage for this.

BTW: JSON will not be the bottleneck at all. It's going to be Jedi itself. We don't have to pick a "fast" protocoll. Jedi is going to be literally a million times slower. Pickle might still be an option though (there are pickle versions aren't there? Or do they collide with python versions?) in case we have or even just pure http/rest.

nakamuray:
> I think this is quite an ok solution. However I don't see what it actually improves besides calling arbitrary python versions (which is great!).

Yes, it only fix version problem, and it is very important for me.
I want my vim which is linked with python 2.7 to get completion from virtualenv I worked on, which is created for python-3.4.
Current jedi.vim state, completion only from standard library of different version of python, is not what I want to get.

> What I would want for an async solution is an async server that is not living within jedi-vim. Such a server should be a part of Jedi! There's a good reason for it: This http server can be used for only within jedi-vim, while it could benefit a few other plugins as well.

No, it's not http server. It uses line based JSON protocol on a pipe.
I put this code on jedi.vim, not jedi itself, because it is vim specific needs.
Other editors, such as emacs and atom, already implement some sort of RPC for their need, because they can't load python modules directly.

> Hence once we're talking about a more general solution, we would need to talk about session handling (because some parts of the API are lazy). How do we keep that state? How do we do garbage collection on such sessions... etc. etc.

Yes, state management is important. So, why do you stick to use HTTP, stateless procotol?
We can use plain TCP or unix domain socket. In this case, connection closed mean "client exited" and safe to release resources created by the client.

> In conclusion: I will not be merging this. It's just - not what we need in the long term. I might merge something that at least contains an async function that makes completion "faster". But I don't see big usage for this.

Sad. Then, I'll try implement it as a separate vim plugin, which override "jedi_vim.jedi" to RPC versioned one.

Anyway, thanks to provide such a great tool!

davidhalter:
> Current jedi.vim state, completion only from standard library of different version of python, is not what I want to get.

It's also not what I want. However, I want a long term solution not something that I have to maintain and throw away again. 

> Other editors, such as emacs and atom, already implement some sort of RPC for their need, because they can't load python modules directly.

True. But this is not the point. It's all about providing a good API on the side of Jedi to avoid 

> Yes, state management is important. So, why do you stick to use HTTP, stateless procotol?
> We can use plain TCP or unix domain socket. In this case, connection closed mean "client exited" and safe to release resources created by the client.

That is true - I'm no expert in that area. I have just suggested HTTP, because it's used a lot at the moment, but a TCP socket would probably be way better.

How do unix domain sockets and TCP sockets compare?

Oh and for the experts: Is it possible to forward TCP sockets over SSH?

PR #512, opened 2015-12-21
Suppress warnings when python3_host_prog or python_host_prog is set

achiku:
- neovim uses Python neovim module where python3_host_prog or python_host_prog is pointing to
- e.g) `let g:python3_host_prog = expand('$HOME') . '/.virtualenvs/py3/bin/python3'`
- If one of those variable is set properly, skip sending warning message
- https://neovim.io/doc/user/nvim_python.html

### DIFF
--- a/autoload/jedi.vim
+++ b/autoload/jedi.vim
@@ -93,7 +93,7 @@ function! s:init_python()
 
         " Add a warning in case the auto-detected version is not available,
         " usually because of a missing neovim module in a VIRTUAL_ENV.
-        if has('nvim')
+        if has('nvim') && (!exists('g:python3_host_prog') && !exists('g:python_host_prog'))
             echohl WarningMsg
             echom "jedi-vim: the detected Python version (".s:def_py.")"
                         \ "is not functional."


davidhalter:
@blueyed I think you know better what this is about since you're using neovim.

autoload/jedi.vim
```diff
@@ -93,7 +93,7 @@ function! s:init_python()
 
         " Add a warning in case the auto-detected version is not available,
         " usually because of a missing neovim module in a VIRTUAL_ENV.
-        if has('nvim')
+        if has('nvim') && (!exists('g:python3_host_prog') && !exists('g:python_host_prog'))
```
blueyed: Shouldn't that be detected / handled earlier, given that it works, with the `has('python')` check in line 90?

achiku: @blueyed @davidhalter Thanks for the comments. I put this condition here rather line 90 since line 90 checks conditions for Vim, not Neovim as it states `if !has('nvim')`.

blueyed: @achiku 
Line 90's right side of the `or` / `||` is used for Neovim.

The logic is: if it is Vim, or (for Neovim) `has('python')` works.

Feel free to fix / improve the flow in this area, but as far as I can see it should work as-is.

What is your use case?
You are using the host prog globals (which are correct), and get a warning?

achiku: @blueyed Oh, you're right. I'm creating repro script but it's taking time a bit, so will reply this weekend!

achiku:
Talking about use case, I want to edit Python file using neovim. Python `neovim` module is installed in virtualenv created by `pyvenv` since I didn't want to install `neovim` module globally. Setting `~/.virtualenvs/py3/bin/python3` to `g:python3_host_prog`, neovim can use `neovim` python module with Python3 interpreter. 

After creating my own project virtualenv by `pyvenv` and workon, right after starting neovim to edit project Python file, it sends warning message as below. 

```
jedi-vim: the detected Python version (2) is not functional. Is the 'neovim' module installed? While jedi-vim will work, it might not use the expected Python path.
```

Although, I'm trying to use Python3 for my project, my default Python in $PATH is still Python2. 

```
≫≫ ll $( which python )
lrwxr-xr-x  1 achiku  staff  37  9 17 13:16 /usr/local/bin/python@ -> ../Cellar/python/2.7.10_2/bin/python2
≫≫ pip freeze | grep nvim
```

That makes `autoload/jedi.vim` `s:def_py` set to 2. Plus, current version of neovim returns 0 when `has('python')` and 1 for `has('python3')`, and skips line 90.

While digging this problem a bit, I think I'm sending wrong p-r....


### neovim version info

```
≫≫ nvim --version
NVIM v0.1.1-122-gf053863 (compiled Dec 26 2015 17:13:30)
```

### system Python info and pip freeze.

```
≫≫ ll $( which python )
lrwxr-xr-x  1 achiku  staff  37  9 17 13:16 /usr/local/bin/python@ -> ../Cellar/python/2.7.10_2/bin/python2
≫≫ pip freeze | grep nvim
≫≫ pip freeze
functools32==3.2.3.post2
futures==3.0.3
gnureadline==6.3.3
greenlet==0.4.9
ipython==3.0.0
isort==4.2.2
jsonschema==2.5.1
mistune==0.7.1
py==1.4.30
pyflakes==0.8.1
python-dateutil==2.4.2
six==1.9.0
stevedore==1.5.0
trollius==2.0
vboxapi==1.0
virtualenv==13.1.0
virtualenv-clone==0.2.6
virtualenvwrapper==4.6.0
wheel==0.24.0
```

### init.vim

`.config/nvim/init.vim` `g:python3_host_prog` settings


```
" =============================
" neobundle settings
" =============================
" Python3 support
let g:python3_host_prog = expand('$HOME') . '/.virtualenvs/py3/bin/python3'
if has('vim_starting')
    set runtimepath+=~/.config/nvim/neobundle.vim/
endif

call neobundle#begin(expand('~/.config/nvim/bundle/'))
```


### virtualenv Python3 (for neovim) info

```
≫≫ workon py3
≫≫ python --version
Python 3.5.0
≫≫ ll $( which python )
lrwxr-xr-x  1 achiku  staff  9 11  4 11:26 /Users/achiku/.virtualenvs/py3/bin/python@ -> python3.5
≫≫ pip freeze
flake8==2.5.0
greenlet==0.4.9
isort==4.2.2
mccabe==0.3.1
msgpack-python==0.4.6
mypy-lang==0.2.0
neovim==0.1.0
pep8==1.5.7
pyflakes==1.0.0
```


When try to edit Python file, neovim displays below warning message.

```
jedi-vim: the detected Python version (2) is not functional. Is the 'neovim' module installed? While jedi-vim will work, it might not use the expected Python path.
```

blueyed:
What about activating your Virtualenv, so that the `python` comes from your Python 3 virtualenv?

Otherwise, you should/could set `g:jedi#force_py_version = 3` also.

achiku:
Python3 virtualenv was activated before opening neovim. It seems like neovim is adding extra paths to `$PATH` environmental variable, and that makes `system('python -c.....)` to use `/usr/local/bin/python`. 

```vim
        " Get default python version from interpreter in $PATH.
        let s:def_py = system('python -c '.shellescape('import sys; sys.stdout.write(str(sys.version_info[0]))'))
        echom s:def_py
        let s:python_path = system('which python')
        echom s:python_path
        let s:os_path = system('echo $PATH')
        echom s:os_path
```

This makes neovim sends messages before starting.

```
2
/usr/local/bin/python^@
/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/X11/bin:/Users/achiku/.virtualenvs/jungle/bin:/Users/achiku/bin:/usr/local/sbin:/Users/achiku/.local/bin:/Users/achiku/.go/1.5.2
/bin:/Users/achiku/.rbenv/shims^@
```

```
(jungle) [~/work/achiku/jungle/jungle] (master|✔)
≫≫ echo $PATH
/Users/achiku/.virtualenvs/jungle/bin:/Users/achiku/bin:/usr/local/bin:/usr/local/sbin:/Users/achiku/.local/bin:/Users/achiku/.go/1.5.2/bin:/Users/achiku/.rbenv/shims:/Users/achiku/.rbenv/shims:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/X11/bin:/Users/achiku/bin:/usr/local/sbin:/Users/achiku/.local/bin:/Users/achiku/.go/1.5.2/bin:/Users/achiku/.rbenv/shims
```

As you said, `g:jedi#force_py_version = 3` works fine, but it disables auto detecting Python version. This issue might not be appropriate in jedi-vim repository?

achiku:
It's been a while and I couldn't come up decent approach to this problem. Could I close this pr?

blueyed:
Yes, let's close this.

achiku:
Thanks!

Issue #516, opened 2016-01-06
g:jedi#popup_select_first=0 selecting last


MikeDacre:
I know bugs with `g:jedi#popup_select_first` have been reported in the past, but I am experiencing a different issue.

`let g:jedi#popup_select_first=1` works as expected, but `let g:jedi#popup_select_first=0` selects the last item, even if all other plugins are disabled.

See this image:

![image](https://cloud.githubusercontent.com/assets/77714/12132695/35e44294-b3d3-11e5-989a-96c4c2d3bf8b.png)

In my vimrc I also have the following completeopt set:

```
autocmd FileType python setlocal completeopt=menuone,longest,preview
```

In my vimrc I also have the following completeopt set:

```
autocmd FileType python setlocal completeopt=menuone,longest,preview
```

My jedi-vim settings:

```
let g:jedi#auto_initialization    = 1
if has( 'python3' )
  let g:jedi#force_py_version     = 3
endif
let g:jedi#popup_on_dot           = 1
let g:jedi#auto_vim_configuration = 1
let g:jedi#popup_select_first     = 0
noremap <leader>gg :call jedi#goto_assignments()<cr>
noremap <leader>gd :call jedi#goto_definitions()<cr>
noremap <leader>rn :call jedi#rename()<cr>
```

Output of `:version`:

```
VIM - Vi IMproved 7.4 (2013 Aug 10, compiled Dec 29 2015 00:04:27)
Included patches: 1-944
Compiled by Arch Linux
Huge version with GTK2 GUI.  Features included (+) or not (-):
+acl             +cmdline_info    +emacs_tags      -hangul_input    +menu            +mouse_xterm     +python3         +syntax          +virtualedit     +xsmp_interact
+arabic          +comments        +eval            +iconv           +mksession       +multi_byte      +quickfix        +tag_binary      +visual          +xterm_clipboard
+autocmd         +conceal         +ex_extra        +insert_expand   +modify_fname    +multi_lang      +reltime         +tag_old_static  +visualextra     -xterm_save
+balloon_eval    +cryptv          +extra_search    +jumplist        +mouse           -mzscheme        +rightleft       -tag_any_white   +viminfo         -xpm
+browse          +cscope          +farsi           +keymap          +mouseshape      +netbeans_intg   +ruby            -tcl             +vreplace
++builtin_terms  +cursorbind      +file_in_path    +langmap         +mouse_dec       +path_extra      +scrollbind      +terminfo        +wildignore
[...]

PR #517, opened 2016-01-06
Fix issue #516

MikeDacre:
Make `g:jedi#popup_select_first=0` work as expected by preventing the
autoselection of the last item when first item not explicitly selected.

Fix for issue [#516](https://github.com/davidhalter/jedi-vim/issues/516)

### DIFF
--- a/autoload/jedi.vim
+++ b/autoload/jedi.vim
@@ -489,6 +489,8 @@ function! jedi#complete_opened(is_popup_on_dot)
         " option is set.
         if g:jedi#popup_select_first && stridx(&completeopt, 'longest') > -1
             return "\<Down>"
+        else
+            return ""
         endif
         if a:is_popup_on_dot
             if &completeopt !~ '\(noinsert\|noselect\)'


MikeDacre:
Note: I have tested this with vim + python2 and vim + python3, it works as expected with both.

autoload/jedi.vim
```diff
@@ -489,6 +489,8 @@ function! jedi#complete_opened(is_popup_on_dot)
         " option is set.
         if g:jedi#popup_select_first && stridx(&completeopt, 'longest') > -1
             return "\<Down>"
+        else
+            return ""
```
davidhalter: However, I'm pretty sure this brings up new issues, because it doesn't respect popup_on_dot.

MikeDacre: @davidhalter Can you clarify how it breaks it? In my hands all combinations of ``popup_on_dot`` and ``popup_select_first`` work as expected. If you can tell me how you think it will cause an issue, I will be very happy to fix it.

blueyed: "popup_on_dot" is only used to trigger the completion (and therefore not relevant here), right?

blueyed: Oh, it's also an argument here (below) and probably for some reason.

blueyed: Have you tested it using the popup_on_dot feature and without "noinsert" or "noselect" in &completeopt?
That's when it's expected to return `<c-p>`, but would return `""` now.

Adding a test for these combinations would be really nice btw.. :)

blueyed:
As noted before this PR by itself disables the code after it.

I've just created https://github.com/davidhalter/jedi-vim/pull/713, which might help with your original issue (#516) maybe, too?

Issue #531, opened 2016-01-18
jedi-vim showing too many popups


strin:
<img width="277" alt="screen shot 2016-01-18 at 12 05 54 pm" src="https://cloud.githubusercontent.com/assets/2183232/12401538/e21214b2-bddb-11e5-8fb2-ee0300b2908d.png">

My .vimrc only has 

```
syntax on                           " syntax highlighing
filetype on                          " try to detect filetypes
filetype plugin indent on    " enable loading indent file for filetype
set rtp+=~/.vim/bundle/vundle/
call vundle#rc()

" jedi-vim: awesome Python autocompletion.
Plugin 'davidhalter/jedi-vim'

```

blueyed:
Confirmed.

Seems to be a Jedi issue though.. `signatures` at https://github.com/davidhalter/jedi-vim/blob/625c568b1249db8c5f1b06a1f9371c06ba2d08b5/jedi_vim.py#L359 is `[<CallSignature: abspath index 0>, <CallSignature: abspath index 0>, <Ca
llSignature: abspath index 0>, <CallSignature: abspath index 0>, <CallSignature: abs
path index 0>]`.

blueyed:
See https://github.com/davidhalter/jedi-vim/pull/532 for a quick attempt to fix this - although I'm not sure this should be handled in jedi-vim - possibly not.

blueyed:
FWIW, getting only 3 (same) signatures currently.

davidhalter:
Fixed since probably Jedi 0.15.

PR #532, opened 2016-01-18
call signatures: make them unique (as strings)

blueyed:
Jedi might return multiple call signatures, where the string
representation is the same.  This patch filters out duplicates.

Fixes https://github.com/davidhalter/jedi-vim/issues/531.

### DIFF
--- a/jedi_vim.py
+++ b/jedi_vim.py
@@ -356,7 +356,11 @@ def show_call_signatures(signatures=()):
         return
 
     if signatures == ():
-        signatures = get_script().call_signatures()
+        sig_strs = []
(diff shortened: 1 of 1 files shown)


davidhalter:
If you want you can merge this. But I kind of disagree with the fix, because it's really a Jedi bug.

davidhalter:
I haven't seen this happening in quite a long time. If it happens again, please report it to Jedi, I'm happy to fix it.

This was definitely an issue a few years ago, but has improved a lot with the inclusion of typeshed and a lot of Jedi fixes.

PR #547, opened 2016-03-18
Improve error messages during init

blueyed:
This will provide more information in case of errors.


davidhalter:
Yeah. That's something we really need. I hope that this helps with all the issues we've been having where jedi-vim wouldn't give you a proper feedback.

Thank you so much for this. It will help the issue tracker a lot :D

PR #548, opened 2016-03-18
jedi#smart_auto_mappings: use `\m` with `search()` for `magic`

blueyed:
With `nomagic` being set (not recommended, but might happen), the
pattern would not work.
This patch sets `\m` explicitly.


davidhalter:
How did you even find this? :D But good find :-) Thanks.

blueyed:
Well, it happened for me.. but I don't know where the `set nomagic` was coming from.

Issue #545, opened 2016-03-11
Rename fails for specific names


spapanik:
Reformat does not  work with specific names

```python
def foo():
    bar = 1
    return bar
```

If I try to rename `bar` to `output` then I get `Jedi did 0 renames!`, although I can rename `bar` to almost any other word.

blueyed:
Fixed in https://github.com/davidhalter/jedi-vim/commit/0d613b5cf976eb23b2be24880d228578039eed59.
Thanks, @mozbugbox!

PR #549, opened 2016-03-20
Fix: rename() kept wrong cursor position #545

mozbugbox:
The cursor would change to something else once in the insert mode. If
the new name was longer than the old one, the Parser couldn't get
the stmt at the wrong cursor position, thus rename failed.

Now save the cursor position before entering the insert mode. Then pass
the saved position to the 2nd call to the rename().

### DIFF
--- a/jedi_vim.py
+++ b/jedi_vim.py
@@ -509,8 +509,10 @@ def rename():
     if not int(vim.eval('a:0')):
         # Need to save the cursor position before insert mode
         cursor = vim.current.window.cursor
+        changenr = vim.eval('changenr()') # track undo tree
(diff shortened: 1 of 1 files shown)


blueyed:
Thanks!

LGTM, but untested.

davidhalter:
I don't see where this argument would be used. Where is it used?

blueyed:
@davidhalter 
What argument?  Should this have been a line comment?

mozbugbox:
The argument is used in the `else` branch as `a:000` and assigned to `cursor`

davidhalter:
Thanks guys :)

Issue #553, opened 2016-04-06
Broken go-to command / "g:jedi#use_splits_not_buffers value is not correct"


yevhen-m:
`g:jedi#use_splits_not_buffers value is not correct, valid options are: winwidth,bottom,top,left,right`

I have this error, everything worked fine yesterday.
I want jedi to open definitions in the current window, this is default behaviour according to docs.

Ubuntu 14.04
NVIM 0.1.3-dev

My jedi settings:
```
let g:jedi#completions_enabled = 0
let g:jedi#auto_initialization = 0
let g:jedi#smart_auto_mappings = 0
let g:jedi#auto_vim_configuration = 0
nnoremap <silent> <leader>g :call jedi#goto()<CR>
```

blueyed:
What is your `g:jedi#use_splits_not_buffers` setting?
It should be `1` by default, and then the warning should not appear..

> I have this error, everything worked fine yesterday.

What has changed?
Have you updated jedi-vim?
Are you using the version from git, or a packaged version?

yevhen-m:
The value is 1, it worked correctly yesterday, but today it is broken.
I've updated plugins today, yes.
Go-to behaviour is strange: it throws me somewhere, but not to the definition. I've seen neovim.api.nvim.NvimError: cursor position outside buffer.
I install packages with vim-plug, so I suppose it's a git version of jedi-vim.

blueyed:
With `1` the following code should not be true?!
https://github.com/davidhalter/jedi-vim/blob/48ad42de86f1c177670967df046f56a20fdfe403/jedi_vim.py#L644

I wonder if it's a type issue..
Please put some `print(type(...))` in there, or figure out in general why the `elif` is true for you.

What Neovim version are you using?  Has this been updated since yesterday?

blueyed:
> I've updated plugins today, yes.

What was the previous version of jedi-vim?  Maybe you can use git-bisect to find where it regressed.

yevhen-m:
I have the value 1 for g:jedi#use_splits_not_buffers, but I get into this elif clause ))). How can it happen?

yevhen-m:
neovim==0.1.7
jedi-vim -- 0d613b5 revision (master)

yevhen-m:
I've commented out 644-658 and now go-to works, but tag stack is not used despite it's enabled by default.

yevhen-m:
[...]

PR #554, opened 2016-04-06
minor: improve error msg with invalid use_splits_not_buffers setting

blueyed:
Ref: https://github.com/davidhalter/jedi-vim/issues/553


davidhalter:
YAY!

PR #555, opened 2016-04-06
jedi#_vim_exceptions: return a string always

blueyed:
Fixes https://github.com/davidhalter/jedi-vim/issues/553.

### DIFF
--- a/autoload/jedi.vim
+++ b/autoload/jedi.vim
@@ -190,6 +190,9 @@ function! jedi#_vim_exceptions(str, is_eval)
     try
         if a:is_eval
             let l:result.result = eval(a:str)
+            if type(l:result.result) != type('')
+                let l:result.result = string(l:result.result)
+            endif
         else
             execute a:str
             let l:result.result = ''


autoload/jedi.vim
```diff
@@ -190,6 +190,9 @@ function! jedi#_vim_exceptions(str, is_eval)
     try
         if a:is_eval
             let l:result.result = eval(a:str)
+            if type(l:result.result) != type('')
+                let l:result.result = string(l:result.result)
+            endif
         else
```
davidhalter: It took me a while to actually understand what happens here. (Because of all the context that is missing in this file). I also haven't read a lot of your ingenious error management :) It looks really good! And I think this is necessary as well.

Would you mind adding a comment? I think it can be confusing, because there's also no comment in `vim_eval` that it returns always a string :)

blueyed: The problem here is that we break list handling etc..

I think there's a regression somewhere else (e.g. old Neovim with new python-neovim) that is causing this.  See the discussion at #553.

blueyed: Another fix would be to cast expected integers using `int()` on the Python side, of course.

davidhalter: You decide. I find it really hard to judge. I think if it's a bug in neovim we should at least document it somehow in our sources, so that it's a little bit easier to follow why some things entered our codebase.

blueyed: @davidhalter 
See #557 and the comment I've amended based on your comment.

blueyed:
Closing this in favour of #557.

PR #556, opened 2016-04-06
Clarify arguments with _catch_exception

blueyed:
The whole int/string conversion with `vim.eval` is confusing enough, use
a boolean for `is_eval`.


davidhalter:
I agree! I actually was confused why it wasn't a boolean in the beginning, when reading your other commits (was probably still my fault, though :)). Thanks!

PR #557, opened 2016-04-07
Cast integers with vim_eval

blueyed:
Cast integers with vim_eval

This makes it clearer/more strict altogether and is future-proof in the
sense that `vim.eval` might return integers for ints in Vim after all.

See #553 for a very noisy discussion of what led to this.
Although it might be a regression in neovim-python after all it is good
to be on the save side.

The bonus here is that we get "notified" in case of a option that is
expected to be a int is a string - although that might also cause some
weird issues in case someone uses "yes" instead of 1 (?!).

Fixes https://github.com/davidhalter/jedi-vim/issues/553.

### DIFF
--- a/jedi_vim.py
+++ b/jedi_vim.py
@@ -154,7 +154,7 @@ def completions():
     row, column = vim.current.window.cursor
     # Clear call signatures in the buffer so they aren't seen by the completer.
     # Call signatures in the command line can stay.
-    if vim_eval("g:jedi#show_call_signatures") == '1':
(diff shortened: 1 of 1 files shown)


davidhalter:
Why would people use `yes` and what is the issue with it?

blueyed:
> Why would people use yes and what is the issue with it (since we're not using bool conversions)?

Removed that from the commit message.

davidhalter:
Makes sense! Thanks!

PR #572, opened 2016-05-24
Improving from-import completion

FredericHamel:
From import completions keep trace of all imported package.

For example:
```python
    from math import abs, 
```
The completion menu (not automatic) will display all other options (excluding abs) like cos, sin, ...
All other component still work.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -49,5 +49,6 @@ Daniel Hahler (@blueyed)
 Dave Honneffer (@pearofducks)
 Bagrat Aznauryan (@n9code)
 Tomoyuki Kashiro (@kashiro)
+Frédéric Hamel (@FredericHamel)
 
 @something are github user names.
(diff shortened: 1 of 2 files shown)


davidhalter:
Can you describe what you have changed? I'm not a 100% sure.

FredericHamel:
I only change completions function in jedi_vim.py.

davidhalter:
But what does the added code do?

FredericHamel:
It fix the completions, for example:
if you type
```python
   from math import cos, 
```
then call completions it will show the other elements of the math package instead of other package like
abc, os, sys, numpy, etc.

From the code:
```python
    from math import cos, cosh, s
```
it  wil create source data like:
```python
    from math import s
```
and tell jedi to complete the possible element in math that start with s.

davidhalter:
Well thanks a lot for the work! I'm sure it works, but this is not a jedi-vim issue, it's a bug in Jedi.

It's closely related to these kinds of issues:
https://github.com/davidhalter/jedi/issues/580

However I'm currently working on this exact issue and need a few weeks.

Sorry for not merging. There's a lot of issues in Jedi, so just feel free to grab one ;-)

PR #582, opened 2016-06-10
Prevent redefinition of InsertLeave in buffers

tweekmonster:
I was testing a plugin (setting/changing `filetype` a lot) and noticed the `autocmd` for closing preview windows was piling up.  As far as I can tell, there isn't a need to rerun the ftplugin after it's already setup.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -49,5 +49,6 @@ Daniel Hahler (@blueyed)
 Dave Honneffer (@pearofducks)
 Bagrat Aznauryan (@n9code)
 Tomoyuki Kashiro (@kashiro)
+Tommy Allen (@tweekmonster)
 
 @something are github user names.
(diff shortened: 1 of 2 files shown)


blueyed:
I think it should use an `augroup` instead:

```vim
augroup jedi_whatever
  au!
  autocmd ....
augroup END
```

This way it will allow for re-sourcing.

@tweekmonster 
Do you want to address / change this in your PR?

tweekmonster:
@blueyed No sweat.  I actually started with an `augroup`, but, I felt like it was overdoing it for some reason.

tweekmonster:
@blueyed I'm done.  I think I should've used the word "duplication" instead of "redefinition".  Let me know if you want me to commit a new message.

blueyed:
Thanks!

I've adjusted the commit message accordingly.

davidhalter:
@blueyed @tweekmonster This is not working right now. If you remove the augroup, it will be removed for all buffers (since there's multiple autocmds (for every buffer) in that group. How could we fix this?

tweekmonster:
Ah, I'll submit a PR.  It needs to be changed to `autocmd! * <buffer>`.

Issue #560, opened 2016-04-16
"'jedi' has no attribute __version__" on system with multiple versions of Python


foxbunny:
After installing jedi-vim from master branch (@a0c326, using pathogen bundle), the following message is shown when opening python files:

```
Could not load jedi python module: module 'jedi' has no attribute '__version__'
```

The system (Arch Linux) has two versions of Python (2.7 and 3.x, where 3.x is default). Jedi library was installed for Python 2.7 only. The error goes away if jedi library is installed under Python 3. As an aside, rope-vim has no trouble using the version of the library installed only for 2.7.

davidhalter:
Probably VIM on arch comes with Python 3. Just install Jedi with Python 3. Or use the git submodule --recursive stuff to initialize the submodule.

davidhalter:
Don't know hot to help you more than the above. Let me know if you still have questions.

blueyed:
After all, this error seems to appear if the `jedi` folder is empty / not checked out.
On Arch Linux you can install `python-jedi` and it will be used (even though there is an empty `jedi` folder in `jedi-vim`).
In general I recommend using the `jedi-vim` repository from GitHub and to make sure the submodule is checked out there.
See #602.

hemna:
I'm still getting this error on ubuntu 16.10 with both python 2.7 and python 3 installed.

I have a complete jedi-vim checkout from git in my ~/.vim/bundle dir.

blueyed:
@hemna 
Please check `:JediDebugInfo` and consider creating a new issue, if it does not help to solve it.

EliazBobadilla:
:(

PR #602, opened 2016-08-03
Improve error msg when the jedi submodule fails to load

blueyed:
Fixes https://github.com/davidhalter/jedi-vim/issues/560.

This PR also improves the README.


davidhalter:
I like. Thanks for improving it!

PR #603, opened 2016-08-06
Add <Plug>(jedi-goto) mapping to allow lazy loading for vim-plug

pryg-skok:
Hello,

For now jedi-vim can't be lazy loaded via plugin managers as vim-plug. But with <Plug> mapping it became possible. So I've add such mapping and now it's possible to load plugin if I try to use jedi#goto() function.

What do use think?
Am I do it in right way?
What I should to add? What functions remap to <Plug>?

### DIFF
--- a/autoload/jedi.vim
+++ b/autoload/jedi.vim
@@ -512,6 +512,8 @@ function! jedi#smart_auto_mappings()
 endfunction
 
 
+nnoremap <silent> <Plug>(jedi-goto) :<C-u>call jedi#goto()<CR>
+
 "PythonJedi jedi_vim.jedi.set_debug_function(jedi_vim.print_to_stdout, speed=True, warnings=False, notices=False)
 "PythonJedi jedi_vim.jedi.set_debug_function(jedi_vim.print_to_stdout)
 


davidhalter:
Well, why would you need that? We already lazy load those functions?!

blueyed:
Yes, especially with `g:jedi#auto_vim_configuration = 0` it is very lazy already.

@pryg-skok 
How does your vim-plug config look with that?
Wouldn't you need more mappings also?

I think having `<Plug>` mappings in general is a good thing though.

pryg-skok:
My config for jedi-vim was simple:
`Plug 'davidhalter/jedi-vim'`

and now I made small modifications:
```
Plug 'pryg-skok/jedi-vim', {'on': '<Plug>(jedi-goto)'}
au FileType python nmap <Leader>d <Plug>(jedi-goto)  " jedi-vim plugin loaded only at first time I really need it
```

@davidhalter I've looked into my vim startup time log and that what I see with simple jedi-vim plug loading:
```
257.962  001.536  001.445: sourcing /Users/fresh/.vim/plugged/jedi-vim/plugin/jedi.vim
1041.799  439.438  439.438: sourcing /Users/fresh/.vim/plugged/jedi-vim/autoload/jedi.vim
1042.640  442.724  003.286: sourcing /Users/fresh/.vim/plugged/jedi-vim/ftplugin/python/jedi.vim
1046.546  000.199  000.199: sourcing /Users/fresh/.vim/plugged/jedi-vim/after/ftplugin/python/jedi.vim
1068.736  000.454  000.454: sourcing /Users/fresh/.vim/plugged/jedi-vim/after/syntax/python.vim
```

And if I make lazy loading with vim-plug it's even don't load while startup! No single line in startup log for jedi-vim.

@blueyed with option g:jedi#auto_vim_configuration = 0 it seems plugin became more lazy but still takes 235ms of startup time - I think it's bad: it must be possible to reduce startup time.

```
099.433  000.054  000.054: sourcing /Users/fresh/.vim/plugged/jedi-vim/plugin/jedi.vim
556.281  235.037  235.037: sourcing /Users/fresh/.vim/plugged/jedi-vim/autoload/jedi.vim
556.839  235.978  000.941: sourcing /Users/fresh/.vim/plugged/jedi-vim/ftplugin/python/jedi.vim
560.095  000.249  000.249: sourcing /Users/fresh/.vim/plugged/jedi-vim/after/ftplugin/python/jedi.vim
576.220  000.200  000.200: sourcing /Users/fresh/.vim/plugged/jedi-vim/after/syntax/python.vim
```

davidhalter:
It takes about 75 ms on my laptop to load jedi-vim. I think this is ok. This only happens when you open Python files and I guess almost all of it is due to `autoload.vim` parsing or executing `s:init_python`. It's the only thing that is actually executed on load. It could be done lazily as well, but I'm not even sure we want that. `s:init_python` loads some Python libraries (including Jedi).

```
182.413  072.568  072.568: sourcing /home/dave/.vim/bundle/jedi-vim/autoload/jedi.vim
182.875  073.079  000.511: sourcing /home/dave/.vim/bundle/jedi-vim/ftplugin/python/jedi.vim
183.173  000.073  000.073: sourcing /home/dave/.vim/ftplugin/python.vim         
183.452  000.172  000.172: sourcing /usr/share/vim/vim74/ftplugin/python.vim    
183.807  000.089  000.089: sourcing /home/dave/.vim/bundle/jedi-vim/after/ftplugin/python/jedi.vim
184.764  000.629  000.629: sourcing /usr/share/vim/vim74/syntax/python.vim      
184.988  000.126  000.126: sourcing /home/dave/.vim/bundle/jedi-vim/after/syntax/python.vim
```

pryg-skok:
> It takes about 75 ms on my laptop to load jedi-vim. I think this is ok. 
Well, I think it depends on performance of your laptop. As you can see it takes 3 times longer on my laptop.

> This only happens when you open Python files
Exactly! And this is the problem, because many users include me open and close files scores of times while working on some project in vim. It's good if vim takes as small startuptime as he can.

davidhalter:
Well, as I said it would be possible to delay the python initialization. I wouldn't mind that. However the solution you've presented doesn't solve anything and I'm therefore closing.

If somebody is interested in doing this, feel free and PR. I won't.

pryg-skok:
Ok, understand. It will be good if project has `<Plug>` mappings - while it's just good, it's also provide specific possibility to on-demand plugin loading.

davidhalter:
Why would that be?

pryg-skok:
Because it's simple clean plugin interface exposed to user.
http://whileimautomaton.net/2008/09/27022735

davidhalter:
While that's obviously cool, it's a one way ticket. It doesn't eliminate all the other settings and would only make jedi-vim less complicated if that's the general way of defining functions.

However since Jedi is being used by far more packet managers than Plug (like pathogen, vam and vundle) this is clearly not feasible.

pryg-skok:
Well, you maintainer - your rules. But why it can affect to other packet managers: it's just an api, that can be exposed or not. If it's exposed - it's good - it can be used.

blueyed:
I agree with @pryg-skok here: the same lazy-loading approach can be used with Neobundle (obsolete) and other plugin managers, too.
But I still think we should expose more/all bindings in the same way.

I think it is really nice to not boot Jedi when only opening a Python file, but only when it actually gets used.

PR #606, opened 2016-08-13
Do not set the '`' mark in goto

blueyed:
This was meant as a hack to make the jumplist work (7520290), but it
works without it.  Maybe the underlying code to go to the definition has
been changed/improved.
Tested both in Vim (7.4.2143) and Neovim.


davidhalter:
Thanks!

PR #608, opened 2016-08-15
Add an ISSUE_TEMPLATE file for GitHub

blueyed:


davidhalter:
I think it's good to add it. I'm not sure if it's not too much already. There's basically 3 points that are overlapping to a certain degree: expected/actual/reproduce. Maybe we could combine those into "Actual Issue" and "Steps to reproduce"? But it's really minor and you can also merge it.

However I like the version stuff very much. I think that helps us a lot!
Maybe we should also add a `:messages` and `:scriptnames`?

The first for potential debugging information and the second one for plugin compatibility issues?

blueyed:
Updated.

davidhalter:
Nice!

blueyed:
This should have been squash-merged (into a single commit), either by updating the PR or using GitHub's squash-merge option with the merge button.
"fixup!"-commits are not meant to be merged into long-lasting branches.  See also https://github.com/marcgibbons/django-rest-swagger/pull/523#issuecomment-240541259.

davidhalter:
Sorry :) Next time I will try!

PR #619, opened 2016-09-25
Reset python init status when autoload/jedi.vim is resourced

blueyed:


davidhalter:
Why would we want that? Doesn't that bring more problems than it solves?

blueyed:
@davidhalter 
I found it useful while improving the `JediDebugInfo` and think that it is good practice in general.
I think all entrypoints should be covered by a call to the `jedi#init_python` method.

Usually you do not re-source a file (and you need to unlet the `loaded` guards, which vim-scriptease is unseful for), but then it's good to reset it's state after all.

davidhalter:
You're the expert ;)

davidhalter:
BTW It looks horrible how much we have to do to initialize Python. :cry: :disappointed:

PR #620, opened 2016-09-25
Hook into Neovim's :CheckHealth with jedi#debug_info

blueyed:
See https://github.com/neovim/neovim/blob/master/runtime/doc/pi_health.txt.

Example output for me:
```
…
health#jedi#check
================================================================================
## jedi
Using Python version: 3
 - sys.version: 3.5.2 (default, Jun 28 2016, 08:46:01), [GCC 6.1.1 20160602]
 - site module: /usr/lib/python3.5/site.py
Jedi path: /home/user/.dotfiles/vim/neobundles/jedi/jedi/jedi/__init__.py
Jedi version: 0.10.0
jedi-vim git version: 0.8.0-42-g7c7c376-dirty
jedi git submodule status: +be2a97cd36ce1b7d6f587ef122efc3433b1e8ae2 jedi (v0.9.0-533-gbe2a97c)
…
```

### DIFF
--- a/autoload/health/jedi.vim
+++ b/autoload/health/jedi.vim
@@ -1,0 +1,4 @@
+function! health#jedi#check() abort
+  call health#report_start('jedi')
+  silent call jedi#debug_info()
+endfunction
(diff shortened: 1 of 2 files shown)


davidhalter:
LGTM. I cannot really reason about this, because I'm not using neovim, but as long as you maintain this piece of code I think it makes sense.

Issue #623, opened 2016-09-29
client/server architecture


dholth:
pythonVScode's completions serializer https://github.com/DonJayamanne/pythonVSCode/blob/master/pythonFiles/completion.py

blueyed:
Please elaborate.

dholth:
It should read separate Jedi Vim into a client and server so the completions can run under a different python than that embedded in vim
On Sep 29, 2016 18:03, Daniel Hahler <notifications@github.com> wrote:Please elaborate.

blueyed:
That is what https://github.com/zchee/deoplete-jedi does and (at least for me) out of the scope for this plugin.

dholth:
Looking for python 2 and pypy completions from a Python 3 vim...
On Sep 29, 2016 18:29, Daniel Hahler <notifications@github.com> wrote:That is what https://github.com/zchee/deoplete-jedi does and (at least for me) out of the scope for this plugin.

blueyed:
Well, you can use Neovim and deoplete/deoplete-jedi then.

But this is not really a support channel.
Closing.

PR #624, opened 2016-10-01
prototype to call jedi completion server in subprocess

dholth:
Call jedi in a subprocess, executed with python on PATH.

This is a very messy prototype of calling Jedi in a subprocess, with less code running in the actual vim-embedded Python. PR for discussion only. I hope it is enough to demonstrate that doing a process split is not a big deal. Sorry that the vendored attrs library adds noise to the diff.

I wanted this to be able to use embedded-Python-3-only vim with Python 2 and importantly PyPy projects. I should probably just recompile vim myself, but I tried this.

The way it works is that jedi_client.py mocks the impressively small jedi API needed by jedi-vim. When you make a call it sends a line of JSON to the server, just a subprocess, which calls real jedi and writes a single line of JSON as the response. Just completions and signatures implemented so far.

If this was developed further then jedi-vim could stay the same, except that a configuration flag imports from jedi_client instead of jedi.

### DIFF
--- a/__main__.py
+++ b/__main__.py
@@ -1,0 +1,13 @@
+"""
+Run the completion server.
+"""
+
+from os import path
+import sys
+
+sys.path.append(path.join(path.dirname(__file__), 'jedi'))
+
+import completion
+
+if __name__ == '__main__':
+    completion.JediCompletion().watch()
--- a/attr/__init__.py
+++ b/attr/__init__.py
@@ -1,0 +1,64 @@
+from __future__ import absolute_import, division, print_function
+
+from ._funcs import (
+    asdict,
+    assoc,
+    has,
+)
+from ._make import (
+    Attribute,
+    Factory,
+    NOTHING,
+    attr,
+    attributes,
+    fields,
+    make_class,
+    validate,
+)
+from ._config import (
+    get_run_validators,
+    set_run_validators,
+)
+from . import exceptions
+from . import filters
+from . import validators
+
+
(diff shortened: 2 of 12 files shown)


blueyed:
Thanks!

But as stated in https://github.com/davidhalter/jedi-vim/issues/623#issuecomment-250610500 I consider this to be out of scope for this plugin, and would assume that @davidhalter thinks similar.
There are frameworks for doing more advanced completion (using a server architecture) already (deoplete and YouCompleteMe at least).

dholth:
I do know about those other projects, particularly I'm a fan of YCM, but it had stopped working in my vim. The other option is switching to a different editor, probably a good idea, but not convenient.

Even when YCM is running it is still nice to be able to run a very simple (jedi-vim) plugin, especially for pypy, and to be able to have the few jedi-vim features that aren't in YCM.

This idea might have more in common with jedi (not jedi-vim) https://github.com/davidhalter/jedi/issues/385 and make more sense there.

davidhalter:
I agree with @blueyed here. There are other plugins that do this if you really need it now.

I think long-term a client/server architecture should be part of Jedi and not jedi-vim. So I'm closing this one. https://github.com/davidhalter/jedi/issues/385

davidhalter:
Thanks for your effort, though! I'm really not against these kind of big pull requests. It's just that the issue I mentioned (davidhalter/jedi#385) has been on my mind for a long time. I think we should really go with that.

PR #625, opened 2016-10-06
Add JediClearCache command

blueyed:
This can be used to work around issues like
https://github.com/davidhalter/jedi/issues/784.

Not sure about the interface and/or the methods it calls - to be discussed.

### DIFF
--- a/autoload/jedi.vim
+++ b/autoload/jedi.vim
@@ -266,6 +266,13 @@ function! jedi#py_import_completions(argl, cmdl, pos)
     PythonJedi jedi_vim.py_import_completions()
 endfun
 
+function! jedi#clear_cache(bang)
+    PythonJedi jedi_vim.jedi.cache.clear_time_caches(True)
+    if a:bang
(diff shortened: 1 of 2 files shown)


davidhalter:
Should I already merge this one (since Jedi is not fixed)?

blueyed:
Yes.
Should it be documented?

davidhalter:
I think we should not document it, yet. There's a chance that it's not necessary anymore with the new parser. Ok? It's also not part of the Jedi API.

blueyed:
Makes sense.

The API being used here is OK though?
Couldn't it become official for Jedi? (apart from this issue)

davidhalter:
> Couldn't it become official for Jedi?

That really depends. IMO the user doesn't want to deal with caches. They should not get corrupted in the first place - so this feels more like a workaround. I know that doesn't really sound good at the moment, but I'm really commited to fixing the existing flaws (having spent a lot of time actually removing problem parts in the last few months).

PR #635, opened 2016-11-16
Update jedi submodule to current master

blueyed:
Comparison: https://github.com/davidhalter/jedi/compare/995a6531225ba0b65e1ff863d97e5404d989047b...2ba78ab725f1e02dfef8bc50b0204cf656e8ee23.

Just suggesting it - feel free to reject it.

I thought it would be good to get some of the fixes into jedi-vim officially.

PR #636, opened 2016-11-18
bugfix for goto(get_script)

chenfjm:
My document structure：
|--A
|----__init__.py
|----a.py
|--B
|----__init__.py
|----b.py
add file content like:
a.py:
`def funa():
    pass
`
b.py:
`from a.aa import funa`

When my corsor on funa in b.py, I can't jump to funa() in a.py .
After change  to `return jedi.Script(source, row, column, buf_path.split("/")[-1], encoding)` i can jump !!!

### DIFF
--- a/jedi_vim.py
+++ b/jedi_vim.py
@@ -149,8 +149,10 @@ def get_script(source=None, column=None):
     if column is None:
         column = vim.current.window.cursor[1]
     buf_path = vim.current.buffer.name
+    if buf_path:
+        buf_path = buf_path.split("/")[-1]
     encoding = vim_eval('&encoding') or 'latin1'
-    return jedi.Script(source, row, column, buf_path.split("/")[-1], encoding)
+    return jedi.Script(source, row, column, buf_path, encoding)
 
 
 @_check_jedi_availability(show_error=False)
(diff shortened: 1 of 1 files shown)


chenfjm:
fix travis check

blueyed:
I think Travis is broken since https://github.com/davidhalter/jedi-vim/pull/635.  Please ignore it for now.

davidhalter:
Change your document structure. I think your imports should include `A` and `B` (plus it feels like a strange problem in general).

The name your talking about is simply the path to the file. I don't know why we should change that. There might be an issue that you've stumbled upon, but it's most likely going to be within the Jedi source.

chenfjm:
Sorry, I made a mistake, the b.py is `from A.a import funa`.

 I just view the Jedi source, in jedi/api/__init__.py, Class Script has two attributes self.path and self._path . When I change self.path to absolute path It's also can jump. I think there is some
bugs about self.path(function: _get_module()). 

Maybe you can fix the bug more simple. I'm expecting .

blueyed:
@chenfjm 
Please report it for the Jedi project, and then reference it here.

chenfjm:
Ok, Thanks.

PR #640, opened 2016-12-04
Change default g:jedi#call_signature_escape to avoid collision

Aulddays:
Syntax highlighting on chars in g:jedi#call_signature_escape may
break the funciton parameter hint from jedi-vim. Change its default
value to non python syntax symbols to reduce chance of collision

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -50,5 +50,6 @@ Dave Honneffer (@pearofducks)
 Bagrat Aznauryan (@n9code)
 Tomoyuki Kashiro (@kashiro)
 Tommy Allen (@tweekmonster)
+Mingliang (@Aulddays)
 
 @something are github user names.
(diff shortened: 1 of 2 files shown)


davidhalter:
Thanks mate!

blueyed:
@Aulddays 
Did it ever work for you after all?

I wonder since there were a lot of issues because of this.

Aulddays:
The setting was overwritten in my .vimrc, so the modification did not really affect my environment...
I'm terribly sorry for that. I should have tested the commit with more caution.

blueyed:
@Aulddays 
I see, not much harm was caused after all.. ;)
It surprised me however more that there were so many duplicate issues reported, and nobody looked into fixing it really - which is strange, since this was error in the Python code, and this plugin is used by Python devs..

davidhalter:
@blueyed My words.

PR #643, opened 2016-12-08
automatic python version check is made async in neovim.

purpleP:
The code is a bit messy, because of my lack of knowledge of vimscript. The problem is that the python version check which was done via `system` call have taken previously half a second which is more or less equal to the time that takes loading everything else including.

With neovim jobcontrol this time reduced to about 80 ms (probably could be improved more, but good enough for me).

The basic idea here is to turn normal code into continuation passing style (which is the only option, because AFAIK vimscript doesn't have any other means for asynchronous code.

You could add vim8 support to that and make proper continuation passing. Right now I've just hardcoded what function to execute if python version check script returned something `s:next_action`.

### DIFF
--- a/autoload/jedi.vim
+++ b/autoload/jedi.vim
@@ -120,6 +120,26 @@ function! jedi#reinit_python()
     call jedi#init_python()
 endfunction
 
+fun! s:JobHandler(jobid, data, event)
+    if a:event == 'stdout'
+        let s:def_py = a:data[0]
+        if &verbose
+            echom "jedi-vim: auto-detected Python: ".s:def_py
+        endif
+    elseif a:event == 'stderr'
+        if !exists("g:jedi#squelch_py_warning")
+            echohl WarningMsg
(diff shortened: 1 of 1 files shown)


davidhalter:
@blueyed  What do you think?

IMO this is a bit too much code for this and we don't really need it? But I'm also not using neovim.

autoload/jedi.vim
```diff
@@ -63,6 +63,7 @@ function! s:init_python()
         try
             return jedi#force_py_version(g:jedi#force_py_version)
         catch
+            echom "error"
```
blueyed: Left-over?!

davidhalter:
@blueyed Fully agree.

blueyed:
Still like the idea, might pick it up after https://github.com/davidhalter/jedi-vim/pull/697 maybe.

blueyed:
This is not necessary anymore, since we do not call `python` in jedi-vim anymore.

PR #650, opened 2017-01-02
Update jedi

fsouza:
Picking up support for Python 3.6


davidhalter:
Nope. This won't happen for now. Sorry. Python 3.6 support is not yet here for Jedi. There's some stuff we need to do first like understanding f"" literals etc.

blueyed:
@davidhalter 
jedi-vim is currently failing hard (silent, errors on stderr; IIRC) on Python 3.6.

I think we should update the Jedi submodule to master? (basically reopening and merging this PR=

davidhalter:
I don't like it, but we can do it. You have to press that button however :package: 

It's not like the bug is solved on the pip version (that most people are using)...

blueyed:
@fsouza 
Thanks!

@davidhalter 
> It's not like the bug is solved on the pip version (that most people are using)...

I wonder why that it different from normal plugins?!  (where people tend to use `master`)

blueyed:
I would suggest cherry-picking bugfixes from `dev` and then releasing a new version maybe?!

davidhalter:
> I would suggest cherry-picking bugfixes from dev and then releasing a new version maybe?!

That's entirely not possible anymore. It's a ~ 230 commit difference and the core of Jedi (type inference) has completely been rewritten. However being really close to seeing the end of the tunnel again, it's not a big issue anymore.

> I wonder why that it different from normal plugins?! (where people tend to use master)

Not one hundred percent sure what you mean, but I'll still try. `master` used to be in sync with the latest release. But I have realized that most projects would use master as a reasonably stable branch to be tested against (like Python itself). I have since started to do the same. I think not pinning the version/commit in a plugin (just linking towards master) is just asking for a lot of issues.

My goal is to synchronize them back up again and do a lot more releases than I used to. The problem was just bad that the test coverage and the code itself was not nearly good enough to be deployable without testing it a lot. It's gotten a lot better in the past year and I think I'm finally at a point where I can do a release at least every three months and I would eventually like to do one every month or so. Right now it's just still too much work.

PR #658, opened 2017-01-14
Improve :JediDebugInfo: verify that JediPython works / sys_path

blueyed:


davidhalter:
Looks good! Thank you for all the effort you put into better debugging.

PR #659, opened 2017-01-25
initialize.py: handle reloading of the plugin

blueyed:
Do not keep adding jedi_vim_path to sys.path.


davidhalter:
IMO we should drop the jedi-vim path as well. It doesn't make a lot of sense to keep it. This would be my preferred fix.

blueyed:
I think it is used for the debug info, but am not sure currently.

blueyed:
Merging this as-is for now.

PR #660, opened 2017-01-25
jedi#debug_info: output non-default settings

blueyed:


davidhalter:
Good idea!

blueyed:
Thanks!
Adopted from https://github.com/neomake/neomake.

PR #662, opened 2017-01-27
Style fixes (via vint)

blueyed:
This also adds a augroup for `jedi#add_goto_window`.

### DIFF
--- a/after/ftplugin/python/jedi.vim
+++ b/after/ftplugin/python/jedi.vim
@@ -9,15 +9,15 @@ if g:jedi#auto_initialization
         setlocal omnifunc=jedi#completions
 
         " map ctrl+space for autocompletion
-        if g:jedi#completions_command == "<C-Space>"
+        if g:jedi#completions_command ==# '<C-Space>'
             " In terminals, <C-Space> sometimes equals <Nul>.
             imap <buffer> <Nul> <C-Space>
             smap <buffer> <Nul> <C-Space>
         endif
-        if g:jedi#completions_command != ""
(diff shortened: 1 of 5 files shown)


autoload/jedi.vim
```diff
@@ -437,7 +442,7 @@ function! jedi#show_call_signatures()
 endfunction
 
 
-function! jedi#clear_call_signatures()
+function! jedi#clear_call_signatures() abort
```
davidhalter: Wow. This seems useful. Not exactly sure what the errors are where it would abort, but still.

after/ftplugin/python/jedi.vim
```diff
@@ -9,15 +9,15 @@ if g:jedi#auto_initialization
         setlocal omnifunc=jedi#completions
 
         " map ctrl+space for autocompletion
-        if g:jedi#completions_command == "<C-Space>"
+        if g:jedi#completions_command ==# '<C-Space>'
```
davidhalter: Looks like javascript. Wtf. But I see why one would use it after reading the docs.

davidhalter:
I think this is a good move.

davidhalter:
I have tried to squash merge. Let me know if you don't like it.

blueyed:
Thanks.
I should have kept the 2nd commit separate though (https://github.com/davidhalter/jedi-vim/pull/662/commits/815ca308517c7c79c204e1aadcfefbe3d755e7bc).
Therefore rebasing/meeting would have been better.

PR #665, opened 2017-01-31
Add omnifunc/completeopt to jedi#debug_info

blueyed:


davidhalter:
Yes, I think that's also an important one.

PR #668, opened 2017-02-02
Only remap `.` if both completions and popup_on_dot are enabled.

thaerkh:
Hello,

Current behaviour would otherwise overwrite any custom vimrc mappings on `.`, regardless of whether or not `g:jedi#popup_on_dot` is set to 0 - this commit addresses that issue.

Cheers,
Thaer

### DIFF
--- a/ftplugin/python/jedi.vim
+++ b/ftplugin/python/jedi.vim
@@ -33,7 +33,7 @@ if g:jedi#auto_initialization
         call jedi#configure_call_signatures()
     endif
 
-    if g:jedi#completions_enabled == 1
+    if g:jedi#completions_enabled == 1 && g:jedi#popup_on_dot == 1
         inoremap <silent> <buffer> . .<C-R>=jedi#complete_string(1)<CR>
     endif
 


blueyed:
> any custom vimrc mappings on `.`

We could also detect if there's a mapping already.
OTOH I think this patch is fine as-is.
Thanks!

thaerkh:
Hi Daniel,

To be clear, this is external to the jedi-vim code. The following's a reproducible example:
```
inoremap .  .<C-x><C-o>
let g:jedi#popup_on_dot = 0
```
The ftplugin line in question will effectively do an `inoremap . .` next time I type the character, overwriting what I have above.

Thanks again,
Thaer

blueyed:
Yes, understood (somehow I guess).
What I've meant was to look at `maparg()` to see if `.` is mapped already.

You still have to set `g:jedi#popup_on_dot = 0` with your patch.

davidhalter:
Closing, because there hasn't been a lot of progress here.

PR #669, opened 2017-02-02
Update the Jedi subrepository.

davidhalter:
Jedi 0.10.0 is out. We should update.


blueyed:
My pleasure!
Thanks! 🎆

wookayin:
FYI -- changelog: https://github.com/davidhalter/jedi/blob/master/CHANGELOG.rst#0100-2017-02-03

Issue #672, opened 2017-02-10
can't disable error msg of python provider checking in nvim


laishulu:
### Issue
as a nvim user, every time I start to edit a python file, a error msg is shown to me.

>jedi-vim: the detected Python version (2) is not functional. Is the "neovim" module installed? While jedi-vim will work, it might not use the expected Python path.

so in nvim init file, I disabled python2 by `let g:loaded_python_provider = 1`, then `:echo has('python') ` will produce `0`.

but from `autoload/jedi.vim`,  `nvim` will always fire the error msg even when python2 or python3 is disabled intentionally in nvim.

```
if has('nvim') || (has('python') && has('python3'))
        " Neovim usually has both python providers. Skipping the `has` check
```

### Versions

 - jedi-vim:  0.8.0-69-g0dea756
 - Vim / Neovim:0.1.17
 - Python: python3

chenhouwu:
```
let g:jedi#force_py_version = 3
```
can avoid the error msg.

davidhalter:
Since this code does not exist anymore I'm closing.

PR #673, opened 2017-02-10
fix: can't disable error msg of python provider in nvim

chenhouwu:
https://github.com/davidhalter/jedi-vim/issues/672


chenhouwu:
```
let g:jedi#force_py_version = 3
```
can avoid the error msg.

PR #674, opened 2017-02-10
fix: even in nvim, python2 and python3 providers may not be provided both

laishulu:
#672

### DIFF
--- a/autoload/jedi.vim
+++ b/autoload/jedi.vim
@@ -68,8 +68,7 @@ function! s:init_python() abort
     endif
 
     " Handle "auto" version.
-    if has('nvim') || (has('python') && has('python3'))
-        " Neovim usually has both python providers. Skipping the `has` check
+    if (has('python') && has('python3'))
         " avoids starting both of them.
 
         " Get default python version from interpreter in $PATH.


autoload/jedi.vim
```diff
@@ -68,8 +68,7 @@ function! s:init_python() abort
     endif
 
     " Handle "auto" version.
-    if has('nvim') || (has('python') && has('python3'))
-        " Neovim usually has both python providers. Skipping the `has` check
+    if (has('python') && has('python3'))
```
blueyed: The `has('nvim')` is meant to not trigger the loading of both Python hosts in Neovim.

blueyed: I think you should set `let g:jedi#force_py_version = 3` instead.

blueyed: I think we could improve it here, but I would then expect `let g:loaded_python_provider = 1` for it to be available.
So in this case you should set `let g:loaded_python_provider = 0`, which also would skip loading it, but then can be considered to mean "do not use it".

blueyed:
```diff
diff --git i/autoload/jedi.vim w/autoload/jedi.vim
index a101836..762bddf 100644
--- i/autoload/jedi.vim
+++ w/autoload/jedi.vim
@@ -69,7 +69,9 @@ function! s:init_python() abort
     endif
 
     " Handle "auto" version.
-    if has('nvim') || (has('python') && has('python3'))
+    if (has('nvim') && get(g:, 'loaded_python_provider', 1)
+          \   && get(g:, 'loaded_python3_provider', 1))
+          \ || has('python') && has('python3')
         " Neovim usually has both python providers. Skipping the `has` check
         " avoids starting both of them.
 ```

blueyed:
Can you check if https://github.com/davidhalter/jedi-vim/pull/697 works for you? (especially https://github.com/davidhalter/jedi-vim/pull/697/commits/a6bbb5d5c794cba0414954480f4cb12732aab8ec)

davidhalter:
Python 2 has been removed, I think this PR is therefore obsolete.

PR #680, opened 2017-02-16
Add instructions for manual installation of the plugin

MarounMaroun:


davidhalter:
Hmm I think if we add it to the README we should also add it to the docs. What do you think?

blueyed:
Yes, makes sense.

MarounMaroun:
@blueyed Do you want me to do any other changes for this PR?

blueyed:
I am more in favor to not include those steps at all, or at least as-is: you would not use `~/.vim/bundle`, and it (the destination) should be included in the `git-clone` command for example.
That being said, I think it could be fixed and then get included.
I don't think it should be included in the Vim doc however - then it is installed already.
If you want to update the PR please do so, otherwise I would do it when merging it.

davidhalter:
Decided not to include this, because it is not a very popular request. It is also something that is not regularly tested and might already not be correct anymore with all the changes to jedi-vim, Jedi and VIM/NeoVIM.

IMO we should rather move towards installing the git module *including* its submodules. Rather than all kind of weird steps and recommend the most popular plugin manager. Everything else would probably just not be supported.

davidhalter:
However @MarounMaroun still thanks for trying to get this change included!

PR #683, opened 2017-02-17
Move Travis targets to Makefile

blueyed:
This allows for calling them easily locally.

Downside: puts installation of `vint` into the `script` part of the Travis job.


davidhalter:
I guess that's fine. make is available on pretty much all the linuxes by default, right?

blueyed:
Yes.

PR #697, opened 2017-04-13
[WIP] Improve Python setup/loading

blueyed:


blueyed:
I think we should change the default from "auto" to "detect" (which just prefers py3).
See https://github.com/davidhalter/jedi-vim/pull/697/files#diff-6db9079804106af1d924941ee58e5cd4R480.

Any feedback, before I push the code changes for this?
/cc @purpleP @chenhouwu

davidhalter:
I have no strong opinions on this. Let me know once you're ready to merge and I'll check if everything makes sense.

blueyed:
Tried to rebase this just now - having an unpushed rebase locally already, but there are many conflicts and empty commits by now.  Likely should be closed, or some small improvements getting picked out of it.

davidhalter:
I'm closing, because this makes improvements mostly to the Python 2/3 usage, whereas now we only use Python 3. So it's much simpler and this change is probably not needed.

PR #699, opened 2017-04-17
JediDebugInfo: settings: display full name

blueyed:


davidhalter:
if you like that better :)

Issue #568, opened 2016-05-11
Improve / disable smart import mapping


blueyed:
Given (at least) #524, #559 and #567 we should consider disabling this by default and/or improving it (also with regard to finding it).

You can quite easily find it by searching for "import" in `:h jedi-vim` though.

So, calling out to the reporters (@hjwp @tinruufu @chkur) of the issues mentioned above:
What do you dislike about it?  (it it too slow?  are you keeping to type ahead?)

I could imagine to skip/abort the process in case the user keeps typing and/or only trigger it after e.g. 50ms after the space has been typed.
What do you think?

I've got used to it myself, but can understand that it's surprising - but would still not say to disable it by default (but without knowing how many like this feature).
Let's have a vote in the next comment.

/cc @aliev as the original inventor of this.. :) (in https://github.com/davidhalter/jedi-vim/pull/454).

blueyed:
Please vote up if you like the feature in general, down if not.

blueyed:
Do you have it enabled currently?  (please vote up/down)

blueyed:
Would you like to have it triggered after a configurable timeout of e.g. 50ms by default, so that it would get skipped when typing the <kbd>i</kbd> directly after the space?

blueyed:
I think it will be easier to implement using a callback in Neovim, so here's another question for up/down voting:
Are you using Neovim?

blueyed:
Is having the same timeout as with `g:jedi#show_call_signatures_delay` OK (which is 500ms) by default?
This setting could be renamed then and be applied for both things.

hjwp:
Hi there, to answer the first question, the reason I wanted to switch it off is I couldnt get used to it -- I did try (not very hard admittedly) to train myself, but I always found myself typing the "import" keyword, and finding it appear twice.

hjwp:
regarding the idea of a short delay before the completion starts.  As someone who doesn't want the feature, as a default behaviour i think that would solve my problem, but:  wouldn't that make the experience worse for the people who do like the feature?

In that respect, maybe it's unlike call signatures?  Because that's something I sometimes want to see and sometimes don't want, so I don't mind waiting a bit if, as I'm coding, just now is one of the times i do want to see it.  but the import keyword autocompletion, i think (maybe?) that's something you either want or don't want, forever.

hjwp:
[...]

PR #708, opened 2017-05-19
Import smart_auto_mappings: use a timer if available

blueyed:
Fixes https://github.com/davidhalter/jedi-vim/issues/568.


Flimm:
This seems to disable the autocomplete magic completely. For instance, typing `from os ` in insert mode does nothing, neither does `from os import `.

`+timers` is enabled for me:

```
$ vim --version | grep timer
+channel         -hangul_input    +netbeans_intg   +timers
```

Within vim, running `:echo g:jedi#smart_auto_mappings` displays `1`.

Flimm:
(Ignore my previous deleted comment.)

I like this behaviour better than the existing behaviour. If I type `from os import` without any pauses between keystrokes, I don't end up with `from os import import`, which is good. So I'm excited about this change.  👍 

Ideally, I would like to see two more changes:

- If I type `from os`, I don't want the plugin to automatically type `import` for me, even after a delay. I don't want the plugin to automatically type anything for me. (But by all means, show the pop-up and insert a suggestion once selected.)
- If I type `from os import`, I would like the pop-up to appear. Right now, it only appears if I type `from os` and then wait.

blueyed:
@Flimm 
Thanks for your feedback.

1. > If I type from os, I don't want the plugin to automatically type import for me, even after a delay

The set `g:jedi#smart_auto_mappings = 0`.

2. > If I type from os import, I would like the pop-up to appear. Right now, it only appears if I type from os and then wait.

jedi-vim does no autocompletion while typing, but only on dots (with `g:jedi#popup_on_dot = 1` (and `g:jedi#completions_enabled = 1`).

I will add special handling for `g:jedi#completions_enabled = 0`, where it will only put the `import.
This is meant to be used with deoplete-jedi or other completion systems.

And while we could have something for your use case there also, I think it would require to have a separate setting for this (or use something like `2` etc for the value).
We could handle `g:jedi#smart_auto_mappings` as a dict, with something like
- `type_import`
- `type_import_delay`
- `complete_after_import`
- `complete_after_import_delay`

It might be better for you to just have a custom mapping maybe?

Flimm:
I just tested this on my home Ubuntu 17.04 computer with vim 8.0, and I couldn't get suggestions or autocompletions to work after typing `from os `, neither in 3176b49 or  	d2d77fa , it simply does nothing. It does work with the code in master.  I'll try again from my other computer and respond to your comments.

davidhalter:
Hmm, @Flimm did you try?

Flimm:
It works on my work computer (macOS, vim 8.0), just not on my Ubuntu 17.04 computer (vim 8.0).

blueyed:
Is it Vim 8.0.0000 on both really?

blueyed:
What does not work exactly? You could add some `echom "here …"` to see what is going on by looking at `:mess` later.

blueyed:
Rebased on master for now.

@Flimm 
Would be helpful if you could try/test it again.

Flimm:
Again, it works on my work computer (macOS), but not on my home computer (Ubuntu 17.04, vim 8.0, package version: 2:8.0.0095-1ubuntu3).

Sorry, I realise that's not very useful, I hope to be able to find time to brush up on my vimscript and debug this, but it's not going to be in the next few days. For the record, I really dislike the old behaviour of typing for me without a timeout, I dislike it so much that I prefer this new behaviour, even when it doesn't anything at all on my Ubuntu computer.

blueyed:
I see.

For debugging:
1. check that `:echom has('timers')` displays 1
2. put some `echom string(…)` into the `jedi#smart_auto_mappings` function to figure out what is happening, and use `:mess` to display the output.  Maybe some of the conditions in the timer callback do not match what is expected? (https://github.com/davidhalter/jedi-vim/pull/708/files#diff-fbb97b52e757230b3e1ea93f959c64a2R634)

autoload/jedi.vim
```diff
@@ -568,12 +578,35 @@ function! jedi#complete_opened(autocomplete) abort
...
+                    \ && mode() ==# 'i'
+                    \ && s:import_timer[2][1:2] == [pos[1], pos[2]-1]
+            " Enter characters and start completion.
+            call feedkeys('import ', 'nt')
+            if g:jedi#completions_enabled
+                call feedkeys("\<C-r>=jedi#complete_string(1)\<CR>", 'nt')
+            endif
+        endif
+        return
+    elseif search('\m^\s*from\s\+[A-Za-z0-9._]\{1,50}\%#\s*$', 'bcn', line('.'))
```
blueyed: The search should be done only in the timer callback (more lazy).

PR #710, opened 2017-05-24
popup_on_dot: document that it needs completions_enabled

blueyed:
It would be ignored already with completions_enabled=0.
This documents it and makes the default follow it.


davidhalter:
That makes sense. Thanks! I took some days off of the issue tracker (holidays), hope you didn't mind.

PR #713, opened 2017-06-03
Completeopt fixes

blueyed:
Fixes https://github.com/davidhalter/jedi-vim/issues/706.

PR #714, opened 2017-06-04
auto_vim_configuration: improve check for changed completeopt setting

blueyed:
The user might want to use the default actually.

Taking this out of #713.

### DIFF
--- a/plugin/jedi.vim
+++ b/plugin/jedi.vim
@@ -15,12 +15,11 @@ if get(g:, 'jedi#auto_vim_configuration', 1)
     " jedi-vim really needs, otherwise jedi-vim cannot start.
     filetype plugin on
 
-    " Change completeopt, but only if it has Vim's default value.
-    let s:save_completeopt=&completeopt
-    set completeopt&
-    let s:default_completeopt=&completeopt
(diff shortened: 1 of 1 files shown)


plugin/jedi.vim
```diff
@@ -15,12 +15,11 @@ if get(g:, 'jedi#auto_vim_configuration', 1)
...
 
-    " Change completeopt, but only if it has Vim's default value.
-    let s:save_completeopt=&completeopt
-    set completeopt&
-    let s:default_completeopt=&completeopt
-    let &completeopt=s:save_completeopt
-    if s:default_completeopt == &completeopt
+    " Change completeopt, but only if it was not set already.
+    redir => completeopt
+        silent verb set completeopt?
```
davidhalter: Hmm, I can kind of see what you are doing here. But isn't the `:silent` causing `completeopt` to be always empty? When I use the silent in command line mode, I don't get any output. I haven't tried it with redir, but I would think that it's probably the same?!

blueyed: No, it is needed here - otherwise it is not (only) redirected (IIRC).

davidhalter:
Makes sense then :) Thanks!

Issue #706, opened 2017-05-19
Automatically types an attribute which doesn't exist but is a argument name


Flimm:
### Issue

`jedi-vim` automatically types an attribute which does not exist, but seems to have originated from the signature of an unrelated function.

### Steps to reproduce

Take this example Python file:

```python
import logging

logger = logging.getLogger(__name__)

def example(**kwargs):
    hello = kwargs['hello']
    logger.info("test %s", hello
```

With my cursor in the last line and the last character in insert mode, I type `.`. Unexpectedly, Jedi inserts the characters `msg`, making the line look like this:

```python
    logger.info("test %s", hello.msg
```

It seems to be getting `msg` from the signature of `logger.info` which is `(param msg, param *args, params **kwargs)`.

I'm sorry, I couldn't work out how to use both `minimal.vimrc` and load `jedi-vim` at the same time.

#### Jedi-vim debug information
Using Python version: 3
 - sys.version: `3.6.1 (default, Apr  4 2017, 09:40:21), [GCC 4.2.1 Compatible Apple LLVM 8.1.0 (clang-802.0.38)]`
 - site module: `/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site.py`
Jedi path: `/Users/flimm/dotfiles/default/vim/bundle/jedi/jedi/jedi/__init__.py`
 - version: 0.10.2
 - sys_path:
    - `/Users/flimm/dotfiles/default/vim/bundle/jedi`
    - `/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python36.zip`
    - `/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6`
    - `/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/lib-dynload`
    - `/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages`
    - `_vim_path_`
    - `/Users/flimm/dotfiles/default/vim/bundle/riv/autoload/riv/`
 - jedi-vim git version: 0.8.0-78-g40a02a7
 - jedi git submodule status:  5427b02712828b2875d35b5ee1c8b5e58f820537 jedi (v0.10.2)

##### Settings
```
g:jedi#smart_auto_mappings = 0 (default: 1)


[...]

PR #715, opened 2017-06-04
Refactor completeopt handling

blueyed:
Temporarily set &completeopt before triggering automatic completion,
instead of handling different settings afterwards.
`g:jedi#popup_select_first` is used to use `noinsert` instead of
`noselect`.

Fixes https://github.com/davidhalter/jedi-vim/issues/516.
Fixes https://github.com/davidhalter/jedi-vim/issues/706.


blueyed:
Merged in a46ea1a.

PR #730, opened 2017-08-08
Use Vim's pythonx mechanism

blueyed:
Move jedi_vim.py and the jedi submodule into pythonx, which gets added
to Vim's internal sys.path.  While jedi cannot be imported directly from
there, it still makes sense for consistency.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "jedi"]
-	path = jedi
+	path = pythonx/jedi
 	url = https://github.com/davidhalter/jedi.git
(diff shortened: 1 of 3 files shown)


davidhalter:
Why do you call it pythonx? Just wondering :)

I like the basic idea.

blueyed:
It is for python2 and python3 - the name is defined by Vim / Neovim.
This is based on the help for Neovim (`:h pythonx`), and I've just noticed that the help is different for Vim - but it works as expected there, too.

There's a `'pyxversion'` setting for Vim, where you can specify your preferred version, which we might want to default to, but that would be another PR/change.

PR #731, opened 2017-08-08
Move Python code for jedi#debug_info into pythonx

blueyed:
This allows vimlparser to parse the vim file again, and it is more
convenient to have it in a real Python file anyway.

Small refactoring and minor text changes included.


davidhalter:
Makes sense! Thanks!

PR #733, opened 2017-08-16
README: improve FAQ for slowness, add deoplete-jedi

blueyed:
Ref: https://github.com/davidhalter/jedi/issues/910#issuecomment-322731875


davidhalter:
Makes sense!

PR #739, opened 2017-09-17
Enhance example installation command

ZhukovAlexander:
Make sure to return to the original directory after git clone

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -89,7 +89,7 @@ Example installation command using Pathogen:
 
 .. code-block:: sh
 
-    cd ~/.vim/bundle/ && git clone --recursive https://github.com/davidhalter/jedi-vim.git
+    cd ~/.vim/bundle/ && git clone --recursive https://github.com/davidhalter/jedi-vim.git && cd -
 
 
 Installation with your distribution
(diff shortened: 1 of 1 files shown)


davidhalter:
I just checked and it looks like this is supported by at least dash/bash/zsh. However I'm still not sure this is a good idea. People actually want to be in the directory where they installed something.

I'm new to `cd -`. Did never see it before, even if we don't merge it, thanks for the hint :)

@blueyed You can decide, I'm somewhat indifferent.

ZhukovAlexander:
> I'm new to cd -

That is exactly why I added it :slightly_smiling_face:  I ended up in another directory after installation and didn't remember the exact original path.

README.rst
```diff
@@ -89,7 +89,7 @@ Example installation command using Pathogen:
 
 .. code-block:: sh
 
-    cd ~/.vim/bundle/ && git clone --recursive https://github.com/davidhalter/jedi-vim.git
+    cd ~/.vim/bundle/ && git clone --recursive https://github.com/davidhalter/jedi-vim.git && cd -
```
blueyed: What about `git clone --recursive https://github.com/davidhalter/jedi-vim.git ~/.vim/bundle/jedi-vim` instead?

ZhukovAlexander: Wow, it totally slipped my mind. This is a perfect one.

blueyed:
Thank you!

PR #741, opened 2017-09-25
jedi#show_documentation: do not use quotes around bufname

blueyed:


davidhalter:
Makes probably sense. I have not tested this one, but it will be easy to revert if there are problems.

PR #746, opened 2017-10-12
Fix tests & update jedi to 0.11.0

davidhalter:
I fixed the tests and re-enabled them to travis-ci.

After that I was finally able to update to `jedi` 0.11.0. `parso` is now also a submodule.

There's quite a lot of changes, but it's actually not that bad. The test changes are just fixes, you can ignore those (because they didn't pass before).

The `jedi_vim.py` changes are not as big as it looks. It's mostly removing things that were deprecated in Jedi a while ago and a big dedent.

BTW: I also removed the call signature problems where param was used everywhere. I don't like that change, but the jedi API is not good enough there.

### DIFF
--- a/autoload/jedi.vim
+++ b/autoload/jedi.vim
@@ -569,10 +569,17 @@ function! jedi#complete_string(autocomplete) abort
         set completeopt+=menuone
         set completeopt-=menu
         if &completeopt !~# 'noinsert\|noselect'
-            if g:jedi#popup_select_first
-                set completeopt+=noinsert
+            " Patch 775 introduced noinsert and noselect, previously these
+            " options didn't exist. Setting them in earlier versions results in
+            " errors (E474).
+            if has("patch-7.4-775")
+                if g:jedi#popup_select_first
+                    set completeopt+=noinsert
+                else
+                    set completeopt+=noselect
+                endif
             else
-                set completeopt+=noselect
+                set completeopt+=longest
             endif
         endif
     elseif pumvisible()
--- a/autoload/jedi.vim
+++ b/autoload/jedi.vim
@@ -579,6 +579,10 @@ function! jedi#complete_string(autocomplete) abort
                     set completeopt+=noselect
                 endif
             else
+                " To pass the tests we use this, it seems to get the closest to
+                " the other options. I'm really not sure if this properly
+                " works, but VIM 7.4-775 is already pretty old, so it might not
+                " be a problem anymore in a few years.
                 set completeopt+=longest
             endif
         endif
--- a/autoload/jedi.vim
+++ b/autoload/jedi.vim
@@ -572,7 +572,7 @@ function! jedi#complete_string(autocomplete) abort
             " Patch 775 introduced noinsert and noselect, previously these
             " options didn't exist. Setting them in earlier versions results in
             " errors (E474).
-            if has("patch-7.4-775")
+            if has('patch-7.4-775')
                 if g:jedi#popup_select_first
                     set completeopt+=noinsert
                 else
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "jedi"]
 	path = pythonx/jedi
 	url = https://github.com/davidhalter/jedi.git
+[submodule "pythonx/parso"]
+	path = pythonx/parso
+	url = https://github.com/davidhalter/parso.git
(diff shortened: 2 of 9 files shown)


blueyed:
Awesome!  Will give that a try and review it, hopefully on the weekend.

blueyed:
Here is your manually delivered coverage report:
```
Name                             Stmts   Miss  Cover   Missing
--------------------------------------------------------------
after/ftplugin/python/jedi.vim      11      1    91%   2
after/syntax/python.vim             22      4    82%   2, 26-28
autoload/jedi.vim                  345    189    45%   43-44, 63-66, 78-85, 96-114, 119-120, 130-132, 144-146, 150, 167, 170-172, 179-251, 256-257, 262-267, 284-285, 296, 300, 304, 308, 312, 316, 324, 328, 332, 344-346, 359-366, 382, 396-408, 413-421, 428, 437-438, 443-445, 456, 473, 486, 511, 524-557, 564, 579, 582, 586, 597
ftplugin/python/jedi.vim            26      2    92%   2, 17
plugin/jedi.vim                     25     11    56%   11, 23-31, 35-36
--------------------------------------------------------------
TOTAL                              429    207    52%
```

It needs quite some hacks to make https://github.com/Vimjas/covimerage/ work here, but we could integrate it.
But looking at it showed again that the test setup is really weird: there is no need for `conftest.py`, the tests should all be in a subdir and I would prefer using Vader over vim-spec probably).
Anyway, besides of the moving nothing too relevant for this PR.

autoload/jedi.vim
```diff
@@ -569,10 +569,17 @@ function! jedi#complete_string(autocomplete) abort
...
+            " errors (E474).
+            if has("patch-7.4-775")
+                if g:jedi#popup_select_first
+                    set completeopt+=noinsert
+                else
+                    set completeopt+=noselect
+                endif
             else
-                set completeopt+=noselect
+                set completeopt+=longest
```
blueyed: Not sure about using longest here then instead.. but I currently do not remember what was used before etc.
Could be added to the comment.

davidhalter:
> But looking at it showed again that the test setup is really weird: there is no need for conftest.py, the tests should all be in a subdir and I would prefer using Vader over vim-spec probably).
Anyway, besides of the moving nothing too relevant for this PR.

I agree that it's a weird setup. I didn't want the ruby dependency at the time. I probably won't refactor, because the current setup works (even though it's very very hacky). But if you want to change it, I'll be happy to merge. I'm not at all a fan of it.

autoload/jedi.vim
davidhalter: I just added longest, because the tests pass on travis with it. I'm pretty sure we used it as default back in the days... However the whole `popup_select_first` stuff is older than `noselect` and `noinsert` and I'm therefore not even sure if that stuff is needed anymore, because these two settings were probably exactly created, because of the problems I had with VIM's autocompletion. Do you still think I should add that as a comment?

davidhalter: I added a comment. Let me know if there are other things that are unclear/that you would like to see changed.

postcapital:
parso seems to be re-caching when opening the same library in a new vim session (I'm using sqlalchemy which is big)  so I think cache is not re-used unless in the same vim session which definitely is cached using a fresh git pull with submodules, so I'm back to using jedi 0.10.2 for time being.  How do you debug a vim plugin presuming it's horrendous?

blueyed:
Is `~/.cache/parso` being created / filled?
Try to remove it between restarts and see how it behaves.
You are using the PR/branch then?

blueyed:
You could also try Jedi's API directly to see how caching behaves, but it might not be trivial.

postcapital:
I've discovered my cache is only re-used if I use an external
jedi==0.10.2 and not the master submodule on 0.10.2 which is not
reused so nothing to do with parso (however 0.11.0 doesn't reload the
cache externally either) .  I am using python 2.7, is that something
to do with it apart from that my system should be the same as anyone
else who checks out the latest master with the 0.10.2 jedi submodule.

postcapital:
Sorry I'm getting in a mess here, externally or not is not a problem.
All I can say is that 0.10.0 reloads the cache twice as quickly as
0.10.2 or 0.11.0 which seems to load new cache.

davidhalter:
@postcapital Can you open an issue in the jedi or parso issue tracker? I think if that's really true that belongs there. If you want you can also try to track it down. That sounds pretty bad.

PR #749, opened 2017-10-15
Usages Highlighting

davidhalter:
Based on #746, we should merge that one first.

Basically it's just two changes: Usages get highlighted in the current module and rename at the end of a line works now better.


after/syntax/python.vim
```diff
@@ -30,3 +30,5 @@ if g:jedi#show_call_signatures > 0 && has('conceal')
         hi jediFat term=bold,underline cterm=bold,underline gui=bold,underline ctermbg=0 guibg=#555555
     endif
 endif
+
+hi jediUsage term=NONE cterm=NONE ctermfg=None guifg=None gui=NONE ctermbg=3 guibg=Yellow
```
blueyed: The color (ctermbg=3) does not look good for me.
What about using `hi jediUsage term=standout cterm=standout gui=standout` ?

davidhalter:
Ok I did what you asked for and I actually think that they are improvements.

I only added TextChanged, should I also add TextChangedI?

blueyed:
> should I also add TextChangedI?

Might make sense, but it is triggered rather often.
Maybe `InsertEnter`?

I am not using the feature much, so cannot really tell - but will do so now.. :)

blueyed:
Seems to work quite well as-is, but still not using it much.

I've noticed though that usages for `attr` and `self.attr` do not include the other one with the following example, i.e. it only finds the current usage.

```python
class C():
    attr = 'foo'

    def foo(self):
        self.attr = 'bar'
```

Likely more a Jedi (or parso) issue - using master for both of them.

blueyed:
I think the following patch makes sense (mainly including the source line in the quickfix list).

Let me know if I should push it.

```diff
diff --git c/pythonx/jedi_vim.py i/pythonx/jedi_vim.py
index 37ee076..4b20893 100644
--- c/pythonx/jedi_vim.py
+++ i/pythonx/jedi_vim.py
@@ -288,16 +288,18 @@ def goto(mode="goto"):
         show_goto_multi_results(definitions)
     return definitions
 
+
 def show_goto_multi_results(definitions):
-    # multiple solutions
+    """Create a quickfix list for multiple definitions."""
     lst = []
     for d in definitions:
         if d.in_builtin_module():
             lst.append(dict(text=PythonToVimStr('Builtin ' + d.description)))
         else:
+            text = '[%s] %s' % (d.description, d.get_line_code().strip())
             lst.append(dict(filename=PythonToVimStr(d.module_path),
                             lnum=d.line, col=d.column + 1,
-                            text=PythonToVimStr(d.description)))
+                            text=PythonToVimStr(text)))
     vim_eval('setqflist(%s)' % repr(lst))
     vim_eval('jedi#add_goto_window(' + str(len(lst)) + ')')
 
diff --git c/pythonx/jedi_vim.py i/pythonx/jedi_vim.py
index 37ee076..4b20893 100644
--- c/pythonx/jedi_vim.py
+++ i/pythonx/jedi_vim.py
@@ -288,16 +288,18 @@ def goto(mode="goto"):
         show_goto_multi_results(definitions)
     return definitions
 
+
 def show_goto_multi_results(definitions):
-    # multiple solutions
+    """Create a quickfix list for multiple definitions."""
     lst = []
     for d in definitions:
         if d.in_builtin_module():
             lst.append(dict(text=PythonToVimStr('Builtin ' + d.description)))
         else:
+            text = '[%s] %s' % (d.description, d.get_line_code().strip())
             lst.append(dict(filename=PythonToVimStr(d.module_path),
                             lnum=d.line, col=d.column + 1,
-                            text=PythonToVimStr(d.description)))
+                            text=PythonToVimStr(text)))
     vim_eval('setqflist(%s)' % repr(lst))
     vim_eval('jedi#add_goto_window(' + str(len(lst)) + ')')
 ```

**Comment by davidhalter** [comment]
@blueyed 
Please push the changes. Regarding the issue above with `attr`, I think this is definitely a Jedi issue. You may open a new ticket if you want.

Your changes are very welcome. I think they make sense. If possible I would try to keep the spacing somewhat similar and use maybe a tab instead of a space (don't know if that works properly though in a VIM quickfix window.

**Comment by blueyed** [comment]
Pushed as is now.
A tab could likely be used.
There could also be custom syntax highlighting, to e.g. use a `Comment` style for the code.

**Comment by davidhalter** [comment]
What do you mean with tabs?

**Comment by blueyed** [comment]
> I would try to keep the spacing somewhat similar and use maybe a tab instead of a space

**Comment by blueyed** [comment]
@davidhalter 
I'd say to :sheep: :it: as it is - after resolving conflicts.

**Comment by davidhalter** [comment]
I would like to merge this once a new Jedi release was made. Also a few things should be improved from my side (see comments). I was just to lazy to actually do it.

**Comment by blueyed** [comment]
Rebased (had conflicts, previous HEAD: fbf17d2).

**Comment by blueyed** [comment]
@davidhalter 
Would be good if you could rebase it yourself to check if the outcome is the same.

**Comment by davidhalter** [comment]
Ok. I rebased. It all seems to work. I also added InsertEnter. I think we can merge.

**Comment by blueyed** [comment]
Will try this today myself.

(My fixup should be squashed in - see `git rebase -i --autosquash` (you can enable `rebase.autosquash` by default).  I can push this myself later, depenging on other findings.)

**Comment by blueyed** [comment]
Pushed a minor improvement.

Regarding the jediUsage hlgroup: it uses `cterm=standout gui=standout`, which does not have an effect for me in Neovim, but only Vim.  IIRC this is fixed with some not-yet-merged patch, but we should probably use something here that works in older Vims and Neovims, too.

**Comment by davidhalter** [comment]
I don't care what we use. We can also just switch depending on the version. Just choose whatever. I think old versions are probably fine, I'm usually not using a "new" version, but we should at least get neovim working.

**Comment by davidhalter** [comment]
Finished?

**Comment by blueyed** [comment]
We should use a default for the highlight that works everywhere, and it should also use `hi def` (see 1773837 in master).
I will take care of it.

Can you say something about my relpath changes?  I feel like there might be a better way.

File: pythonx/jedi_vim.py [hunk 2]
```diff
@@ -340,24 +336,65 @@ def goto(mode="goto", no_output=False):
...
+        show_goto_multi_results(definitions)
     return definitions
 
 
+def relpath(path):
+    """Make path relative to cwd if it is below."""
+    abspath = os.path.abspath(path)
+    if abspath.startswith(os.getcwd()):
+        return os.path.relpath(path)
+    return path
```
davidhalter: I like the relpath change a lot.

In the future we could probably start being even more clever and return a relative path if it's only one/two levels outside or something like that, but I'm not even sure that's a good idea, so thumbs up :)

I actually thought about something similar as well, I would probably use relpath everywhere where we use `escape_file_path` (or even in that function), because I generally want VIM to use short paths for displaying a buffer/tab path. Is there some issue that stops us from doing that?

blueyed: I've thought there would be a better function for it already, and/or that the function itself could be better - just came up very quickly with it.
In Vimscript there would be `fnamemodify(…, ':p:.')` - and `":~:."` to have it relative to HOME.

If you like it like it is, then I think we're good.

I somehow dislike that the window closes automatically when leaving it - I often would like to keep it open for later, but it is easy to comment out for me, and also to retrigger it.
However, maybe something to consider?

davidhalter: I would probably leave it for now as it is and wait for feedback. We could also use other keys like `o` (buffer), `t` (tab) to open the buffer at that point and keep the window open.

blueyed:
I think for actual usages the tree_name (?) should not be there additionally, i.e. in the example below the 1st and 3rd entry should not contain `[_set_source_attributes] `:

```
rplugin/python3/deoplete/deoplete.py|213 col 14| [_set_source_attributes] self._set_source_attributes(context)
rplugin/python3/deoplete/deoplete.py|220 col 9| [def _set_source_attributes] def _set_source_attributes(self, context):
rplugin/python3/deoplete/deoplete.py|235 col 18| [_set_source_attributes] self._set_source_attributes(context)
```

Also the second entry could be just "def" maybe? - although that is then duplicated right there.

The point is that the name itself is usually in the source already.

What about the following?

```diff
diff --git i/pythonx/jedi_vim.py w/pythonx/jedi_vim.py
index ccafc74..902689b 100644
--- i/pythonx/jedi_vim.py
+++ w/pythonx/jedi_vim.py
@@ -359,7 +359,15 @@ def show_goto_multi_results(definitions):
             # well.
             lst.append(dict(text=PythonToVimStr(d.description)))
         else:
-            text = '[%s] %s' % (d.description, d.get_line_code().strip())
+            # Uses d.description (rather than d.type) to get e.g. "def" for
+            # "function".
+            code = d.get_line_code().strip()
+            desc = d.description.split(' ')
+            if (len(desc) > 1
+                    and (desc[0] != 'def' or not code.startswith('def'))):
+                text = '[%s] %s' % (desc[0], code)
+            else:
+                text = d.get_line_code().strip()
             filename = relpath(d.module_path)
             lst.append(dict(filename=PythonToVimStr(filename),
                             lnum=d.line, col=d.column + 1,
```

Or we use a short version of `d.type` everywhere, e.g. `stmt` for statement (which is the first and third line).

Sorry for bikeshedding so much - just some random ideas/feedback.

davidhalter:
I'm happy with pretty much every improvement. I can definitely see that the notation we are using currently doesn't have much value.

davidhalter:
I'm not sure if you did already change it, but didn't you want to change the standout/reverse to make it work for you in neovim?

blueyed:
It uses reverse for that reason already - I think I fixed up your commit.

PR #763, opened 2017-12-14
Upgrade jedi and parso

davidhalter:


davidhalter:
Just merged it, let me know if there are problems :)

PR #765, opened 2017-12-28
fix: CheckHealth doesn't work in neovim

naronA:
Hi. I found CheckHealth bug in neovim.
CheckHealth calls `display_debug_info()`.
However, error occured when loading `jedi_vim.jedi.Script('')._evaluator.sys_path`,
because `Evaluator.sys_path` is unavailable variable.

### DIFF
--- a/pythonx/jedi_vim_debug.py
+++ b/pythonx/jedi_vim_debug.py
@@ -31,7 +31,13 @@ def display_debug_info():
             echo("printf(' - version: %s', {0!r})".format(
                 jedi_vim.jedi.__version__))
             echo("' - sys_path:'")
-            for p in jedi_vim.jedi.Script('')._evaluator.project.sys_path:
+
+            script_evaluator = jedi_vim.jedi.Script('')._evaluator
+            try:
+                sys_path = script_evaluator.project.sys_path
+            except AttributeError:
+                sys_path = script_evaluator.sys_path
+            for p in sys_path:
                 echo("printf('    - `%s`', {0!r})".format(p))
     except Exception as e:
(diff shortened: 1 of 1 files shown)


davidhalter:
If this broke, I'm sorry. It will break again once the virtualenv branch is merged. But then we'll probably have a public API to ask for this stuff.

naronA:
Oh I see.
Sorry, I did not check virtualenv branch.
This change will break again certainly.
Thanks!

blueyed:
I think we should re-open and actually try to fix it, by handling the different variants (i.e. the current one, the one from this PR, and the one from the virtualenv branch).

@naronA 
Can you update the PR, please?

davidhalter:
Please don't try to work with the virtualenv branch, yet. It's work in progress and we can integrate it as soon as it's in good shape and the API is final.

As for the rest, I agree that here in this pull request we should try to handle both those variants.

naronA:
Thanks for replies.
Should I new PR when virtualenv branch development is finished?
(I'm Sorry I'm not familiar with github.)

blueyed:
No, I would suggest updating this PR to handle both the old and new "API".
```python
script_evaluator = jedi_vim.jedi.Script('')._evaluator
try:
    sys_path = script_evaluator.project.sys_path:
except AttributeError:
    sys_path = script_evaluator.sys_path
```

naronA:
ahh, I have misunderstood that new API will be added.
I'm sorry for lack of my English proficiency.
I will update my PR.

davidhalter:
Thanks!

PR #769, opened 2018-01-15
Improve tests / failure detection

blueyed:


davidhalter:
I guess this is ok as long as all tests still pass :)

PR #771, opened 2018-01-24
goto: use only definitions with module_path

blueyed:
Works around https://github.com/davidhalter/jedi/issues/1033.

### DIFF
--- a/pythonx/jedi_vim.py
+++ b/pythonx/jedi_vim.py
@@ -304,7 +304,11 @@ def goto(mode="goto", no_output=False):
         return definitions
     if not definitions:
         echo_highlight("Couldn't find any definitions for this.")
-    elif len(definitions) == 1 and mode != "related_name":
+        return definitions
+
+    definitions = [d for d in definitions if d.module_path]
+
+    if len(definitions) == 1 and mode != "related_name":
         d = list(definitions)[0]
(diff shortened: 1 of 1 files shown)


blueyed:
Does this make sense in general?
Even with https://github.com/davidhalter/jedi/issues/1033 being fixed it makes no sense to have definitions without a path here.. am I missing something?

davidhalter:
Well it depends. Let's say you want to find all usages of `math.cos`. Do you want to ignore `math.cos` itself? Shouldn't we list that as a "non-clickable" so that the information is there that this is the definition? IMO it would then look something like this:

```
builtin math|12 col 8| module math
/foobar/api/__init__.py|90 col 21| math
/foobar/api/example.py|104 col 47| math
api/environment.py|1 col 8| module os
```
I think you make a good point, but I'm not really sure what we want here. Ideally in a normal GUI I would probably strive for a way to differentiate between definitions and usages. This is something we could do here as well, but not in such a graphical way I guess.

One thing I should also mention is that if you open vim without a path and then use `:filetype python` you actually don't have a path - and might run into your code change. This might not be an extremely important use case, but it's just important for us to know that we break that.

davidhalter:
This is obsolete in my opinion, feel free to let me know if there is still an issue here.

Issue #748, opened 2017-10-15
README.rst Enhancement Proposal


zwidny:
### Issue
Install Jedi by ReadMe.rst by step, but it does not work correctly.

### Steps to reproduce

I install jedi-vim by README.rst, and I use  Vundle to install jedi-vim.
The doc say nothing to how to install jedi-vim by Vundle,  but say 
```
git clone --recursive https://github.com/davidhalter/jedi-vim.git ~/.vim/bundle/jedi-vim
```
only. so I think Vundle do not need add this line"Plugin 'davidhalter/jedi-vim'"

That make my jedi does not work, and I install by README.rst, so I do not the reason.

### Suggest
Add Vundle install doc. e.g.:

For Vundle,  need add this
```
Plugin 'davidhalter/jedi-vim'
```
to ~/.vimrc

mmequignon:
Hi @zwindy !
Yes, as you noticed, there is no install description for Vundle.
I will create a pull request.

davidhalter:
Thanks for bringing this up. I think we can continue the discussion in #781.

PR #781, opened 2018-02-15
[IMP] README.rst - Add a description for Vundle installation.

mmequignon:
Hi !
As @zwidny noticed in #748, there is no description in the README for Vundle install type.
This pr just adds the missing description.
Would not it be better to add a dedicated subsection for Vundle install type ?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -91,6 +91,14 @@ Example installation command using Pathogen:
 
     git clone --recursive https://github.com/davidhalter/jedi-vim.git ~/.vim/bundle/jedi-vim
 
+Example installation using Vundle:
+
+Add the following line in you `~/.vimrc`
+    
+.. code-block:: vim
+
+    Plugin 'davidhalter/jedi-vim'
+
 
 Installation with your distribution
 -----------------------------------
--- a/README.rst
+++ b/README.rst
@@ -93,7 +93,7 @@ Example installation command using Pathogen:
 
 Example installation using Vundle:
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -91,6 +91,14 @@ Example installation command using Pathogen:
 
     git clone --recursive https://github.com/davidhalter/jedi-vim.git ~/.vim/bundle/jedi-vim
 
+Example installation using Vundle:
+
+Add the following line in you `~/.vimrc`
```
davidhalter: Typo. :) `in your`.

I think `to your` would be better as well, but then again I'm not an English native.

mmequignon: Yes, you're right, it's a typo.
It's fixed.

mmequignon: Concerning the documentation file, it seems already alright.

mmequignon:
I can squash commits before the merge if you are fine with the modifications.

davidhalter:
I can also squash here in the web interface. No problem.

mmequignon:
Well, that's all right for me.
That you for your quick answers !

mmequignon:
> Do you think that it's necessary? @mmequignon
No I don't, it was just an idea.

davidhalter:
I just realized that vundle is already documented here. https://github.com/davidhalter/jedi-vim/blob/master/doc/jedi-vim.txt

I guess in that case it makes sense to just add it to the README.

mmequignon:
Thank you for merging !

PR #801, opened 2018-04-17
get_script: additional_dynamic_modules: add only listed buffers

blueyed:
Unlisted buffers might come e.g. from `set viminfo+=%`, and Jedi would
crash on them.

This could additionally check for them to be readable, but this should
be handled in Jedi itself anyway.


davidhalter:
That makes a lot of sense. Thank you.

PR #806, opened 2018-04-17
Virtualenv changes

davidhalter:
This is mostly needed, because we are now using jedi 0.12.0.

Most of the changes are minor. The bigger one is that force_py_version now just affects the environment version (which is a lot better and works better). You cannot choose between VIM's Python 2/3 anymore.

I think that a more interesting thing will be to use jedi.find_virtualenvs and jedi.find_system_environments. Users should then be allowed to switch to those. But that's material for another pull request and just a new feature. Plus I'm not even sure that I'm going to do that.


davidhalter:
Tests are failing. I will investigate soon.

davidhalter:
It's working now. I forgot to merge... :) Can you please review and merge @blueyed? It fixes the :JediDebugInfo command among other things.

pythonx/jedi_vim.py
```diff
@@ -25,6 +25,15 @@
     ELLIPSIS = u"…"
 
 
+try:
+    sys.prefix = sys.base_prefix
+    sys.exec_prefix = sys.base_exec_prefix
+except AttributeError:
+    # Somehow sys.prefix is set in combination with VIM and virtualenvs.
+    # However the sys path is not affected. Just reset it to the normal value.
+    pass
```
blueyed: What is this needed for?  (in the case of not AttributeError)

pythonx/jedi_vim.py
```diff
@@ -148,6 +157,36 @@ def wrapper(*args, **kwargs):
...
 
 
+last_force_python_error = None
+
+
+def get_environment():
+    global last_force_python_error
+
+    force_python_version = vim_eval("g:jedi#force_py_version")
+    environment = jedi.api.environment.get_cached_default_environment()
```
blueyed: Could be skipped with `!= "auto"` - only needed in case on an exception then.

pythonx/jedi_vim.py
```diff
@@ -148,6 +157,36 @@ def wrapper(*args, **kwargs):
     return func_receiver
 
 
+last_force_python_error = None
+
+
+def get_environment():
+    global last_force_python_error
+
+    force_python_version = vim_eval("g:jedi#force_py_version")
```
blueyed: This could optionally become a buffer-local variable, so that you could use different versions per buffer then.
`vim_eval("get(b:, 'jedi_force_py_version', get(g:, 'jedi#force_py_version')")`

davidhalter: I agree. I just think that this should be part of another pull request.

pythonx/jedi_vim.py
davidhalter: Bad documentation. Sorry. It's at least somewhat better, now.

pythonx/jedi_vim.py
davidhalter: I changed it.

pythonx/jedi_vim.py
```diff
@@ -158,36 +158,39 @@ def wrapper(*args, **kwargs):
...
+
+    vim_force_python_version = vim_eval("g:jedi#force_py_version")
+    if use_cache and vim_force_python_version == current_environment[0]:
+        return current_environment[1]
 
-    force_python_version = vim_eval("g:jedi#force_py_version")
     environment = None
-    if force_python_version != "auto":
+    if vim_force_python_version == "auto":
+        environment = jedi.api.environment.get_cached_default_environment()
```
davidhalter: If we cache anyway, we can just use jedi.get_default_environment() instead. It's the official API, this one is kind of hidden and not really public.

doc/jedi-vim.txt
```diff
@@ -480,21 +480,20 @@ small a horizontal split happens.
...
+let g:jedi#force_py_version = 2
 
-This variable can be switched during runtime using the following function:
-Function: `jedi#force_py_version_switch()`
+By default jedi loads the latest Python version installed on your system that
+can be found.
 
-or set directly using this function, which has the same name as the variable:
-Function: `jedi#force_py_version(py_version)`
+This variable can be switched during runtime.
```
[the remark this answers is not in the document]
davidhalter: I don't understand. What is failing? It probably shouldn't.

davidhalter:
I added two comments, you probably have to show those (`Show outdated`).

davidhalter:
I merged now, because I probably fixed the remaining issues and not merging it caused JediDebugInfo to not work.

I'm happy to do further changes if they are necessary.

Issue #814, opened 2018-05-05
goto: TypeError: an integer is required with ImplicitNSName


blueyed:
Using `goto` on `import jedi` in pythonx/jedi_vim.py causes an error, because it uses `None` for the line and the column of the definition:

Vim:
```
Traceback (most recent call last):
  File "/home/user/.vim/plugged/jedi-vim/pythonx/jedi_vim.py", line 131, in wrapper
    return func(*args, **kwargs)
  File "/home/user/.vim/plugged/jedi-vim/pythonx/jedi_vim.py", line 297, in goto
    vim.current.window.cursor = d.line, d.column
TypeError: an integer is required (got type NoneType)
```

Neovim:
```
Traceback (most recent call last):                    
  File "/home/user/.vim/plugged/jedi-vim/pythonx/jedi_vim.py", line 141, in wrapper
    return func(*args, **kwargs)
  File "/home/user/.vim/plugged/jedi-vim/pythonx/jedi_vim.py", line 345, in goto
    vim.current.window.cursor = d.line, d.column
  File "…/Vcs/neovim-python-client/neovim/api/window.py", line 27, in cursor
    return self.request('nvim_win_set_cursor', pos)
  File "…/Vcs/neovim-python-client/neovim/api/common.py", line 44, in request
    return self._session.request(name, self, *args, **kwargs)
  File "…/Vcs/neovim-python-client/neovim/api/nvim.py", line 170, in request
    res = self._session.request(name, *args, **kwargs)
  File "…/Vcs/neovim-python-client/neovim/msgpack_rpc/session.py", line 100, in re
quest
    raise self.error_wrapper(err)
neovim.api.nvim.NvimError: b'Argument "pos" must be a [row, col] array'
```

> Definition(name=ImplicitNSName(string_name='jedi', start_pos=None), full_name='jedi', description='module jedi')

This appears to happen because Jedi is not installed in the virtualenv (via `$VIRTUAL_ENV`) (but I am using the system Python where it is used from), and then gets picked up as an entry from `pythonx` of the jedi-vim plugin?!  (just a guess)

I guess the fix would be similar to 79ced22.

blueyed:
Could/should ImplicitNSName have some path attribute that could be used here?

E.g. in Vim the directory (…/pythonx/jedi) could be opened then.

This is what I get when using a python where jedi is not installed from the
`pythonx` dir (to simulate what Neovim/Vim are doing for `pythonx`):

```
[...]

PR #815, opened 2018-05-05
WIP: fix goto for namespaces

blueyed:
TODO:

- [ ] should prefer module_path if available
- [ ] should not use private API
- [ ] should not have to `or` `line`/`column`

Ref: https://github.com/davidhalter/jedi-vim/issues/814


davidhalter:
As noted in #814, I feel like this is fine now. 

The only thing that might be left to do is changing the error message so that the type appears: `jedi-vim: Builtin modules cannot be displayed (ftplugin:module ftplugin).`). It could say that it's a namespace, but that's all.

PR #822, opened 2018-05-21
Fix s:init_python

blueyed:
Just check for `has('python3')` / `has('python')`.

Fixes https://github.com/davidhalter/jedi-vim/commit/72a2643f7c0e6f895b7bf61d2b180c42b143aa7b#r28922063.

### DIFF
--- a/autoload/jedi.vim
+++ b/autoload/jedi.vim
@@ -58,47 +58,10 @@ endfor
 let s:script_path = fnameescape(expand('<sfile>:p:h:h'))
 
 function! s:init_python() abort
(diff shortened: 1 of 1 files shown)


blueyed:
We're losing (but already have lost) the ability to enforce/use e.g. `python` (instead of `python3`) on Neovim (where it is more expensive to have multiple interpreters), e,g, when you only want to use a single interpreter and some plugin requires Python 2.

But defaulting to `python3` seems to be good enough, since it should be the default/preferred version in the long run.

davidhalter:
Makes sense!

Issue #833, opened 2018-06-21
Error detected while processing function jedi#init_python


regisz:
### Issue

I get the next error when I open a py file at first:
```
Error detected while processing function jedi#init_python:
line    7:
Error: jedi-vim failed to initialize Python: jedi#setup_python_imports: could not import jedi_vim: SyntaxError: invalid syntax (jedi_vim.py, line 25). (in function jedi#init_python[3]..<SNR>30_init_python[2]..jedi#setup_python_imports, line 30)
```

### Steps to reproduce

Open a python file.

Your template `minimal.vimrc` solution is enough to reproduce the error.

Save it as `minimal.vimrc` in the directory where jedi-vim is installed, `cd` into that directory, and run Vim with `vim -u minimal.vimrc`):

```
set nocompatible

let script_dir = fnamemodify(expand('<sfile>'), ':h')
let &runtimepath .= ','.script_dir.','.script_dir.'/after'

" Put your config changes here.
" let g:jedi#show_call_signatures=1

syntax on
filetype plugin indent on
```

### Output of “:verbose JediDebugInfo”

You should run this in a buffer with filetype "python".
#### Jedi-vim debug information
Using Python version: 3
 - global sys.version: `3.2.3 (default, Mar 25 2017, 10:17:28), [GCC 4.7.2]`
 - global site module: `/usr/lib/python3.2/site.py`
ERROR: jedi_vim is not available: SyntaxError: invalid syntax (jedi_vim.py, line 25)
 - jedi-vim git version: 0.9.0-21-g84a5299c8f
 - jedi git submodule status:  bd7c65d9639c2d451d44c6d135b4f6043a918950 pythonx/jedi (v0.12.0)
 6f385bdba182059e0c46917a414981a6d0894527 pythonx/parso (v0.2.1)

##### Settings
```
  omnifunc=
  completeopt=menuone,longest,preview
        Last set from ~/.vim/bundle/jedi-vim/plugin/jedi.vim
```

#### :version
```

VIM - Vi IMproved 8.1 (2018 May 17, compiled Jun 20 2018 15:20:20)
Included patches: 1-89
Compiled by balage
Huge version without GUI.  Features included (+) or not (-):
+acl               +cmdline_hist      +eval              +job               +mouse             +num64             +scrollbind        +termresponse      +windows
+arabic            +cmdline_info      +ex_extra          +jumplist          -mouseshape        +packages          +signs             +textobjects       +writebackup
[...]

PR #834, opened 2018-06-23
s:init_python: handle g:jedi#force_py_version again

blueyed:
It was removed previously, since it now refers to the environment being
used.

This brings it back for now.

There could be a separate var for it later, and/or it needs changes
later (or rather the environment needs a new variable name, since it
should refer to the executable really) - but this is good for now I think.

Fixes https://github.com/davidhalter/jedi-vim/issues/833

### DIFF
--- a/autoload/jedi.vim
+++ b/autoload/jedi.vim
@@ -57,14 +57,35 @@ endfor
 " ------------------------------------------------------------------------
 let s:script_path = fnameescape(expand('<sfile>:p:h:h'))
 
+" Initialize Python (PythonJedi command).
 function! s:init_python() abort
-    if has('python3')
-        call jedi#setup_python_imports(3)
-    elseif has('python')
-        call jedi#setup_python_imports(2)
+    " Use g:jedi#force_py_version for loading Jedi, or fall back to using
+    " `has()` - preferring Python 3.
+    let loader_version = get(g:, 'jedi#loader_py_version', g:jedi#force_py_version)
+    if loader_version ==# 'auto'
+        if has('python3')
+            let loader_version = 3
+        elseif has('python2')
+            let loader_version = 2
+        else
+            throw 'jedi-vim requires Vim with support for Python 2 or 3.'
+        endif
     else
-        throw 'jedi-vim requires Vim with support for Python 2 or 3.'
+        if loader_version =~# '^3'
+            let loader_version = 3
+        elseif loader_version =~# '^2'
+            let loader_version = 2
+        else
+            if !exists('g:jedi#squelch_py_warning')
+                echohl WarningMsg
+                echom printf("jedi-vim: could not determine Python loader version from 'g:jedi#loader_py_version/g:jedi#force_py_version' (%s), using 3.",
+                            \ loader_version)
+                echohl None
+            endif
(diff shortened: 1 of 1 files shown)


blueyed:
@davidhalter 
What do you think about having the old behavior (only) with `g:jedi#force_py_version`, and adding something like `g:jedi#use_python` (and `b:jedi_use_python`) for specifying the environment then?

The tricky part currently is that you cannot differentiate between the Python to load Jedi and the Python that Jedi uses.

I think having used `force_py_version` for the new environments was wrong.

If you agree, the following should get applied here, too:
```diff
diff --git i/autoload/jedi.vim w/autoload/jedi.vim
index d039e8a..3c1a229 100644
--- i/autoload/jedi.vim
+++ w/autoload/jedi.vim
@@ -70,9 +70,9 @@ function! s:init_python() abort
           throw 'jedi-vim requires Vim with support for Python 2 or 3.'
       endif
     else
-       if g:jedi#force_py_version =~# '^3'
+       if g:jedi#force_py_version ==# '3'
            let py_loader_version = 3
-        elseif g:jedi#force_py_version =~# '^2'
+        elseif g:jedi#force_py_version ==# '2'
             let py_loader_version = 2
         else
             throw printf("jedi-vim: could not determine Python version from '%s' (%s)",
```

davidhalter:
IMO I wouldn't do that. Python 2 is dying soon and I'm probably not going to support it beyond 2020. People should switch. If they still want Python 2, they can just use an older version of everything without bug fixes and stuff.

This PR goes into the same direction, we're making changes, because someone is having issues with Python 3.2. IMO this is just too much maintenance work for little gain. It seems to be working very well with most users and there has been one case where it didn't, but that's really just one case.

So as I said, I'm kind of against it, but if you really want this we can maybe just use a new variable `force_base_python` that we don't really mention loudly in docs. Your call.

davidhalter:
> Re g:jedi#force_py_version: do you want to keep using it for the envs then?

Yes. I think most people just want to specify a Python version and not an environment. I think the environment should be a separate setting if people really want that.

blueyed:
Hmm.
So a new setting for the base Python should be used then (this PR).

`g:jedi#force_py_version` could then support both versions (could be looked up from the envs), or an explicit env (using its executable).  I find the name a bit misleading though, so that might have been a good chance to clean this up.

davidhalter:
If you changed something that was related to my comment, please rebase, the "old" logos are still in this diff.

blueyed:
This allows for `g:jedi#loader_py_version` to override `g:jedi#force_py_version`, but does not document it.

davidhalter:
I have quickly reviewed it and it looks generally good. I generally like the naming.

The point is just, I'm really against re-introducing this. Apart from one guy that runs Python 3.2 (and that's extremely old) - nobody has complained about the lack of this feature. This guy could also just be using Jedi < `0.10.0` with an older jedi-vim version, which should still work.

Since I don't fell like I should just decide this, I'll let you. You did the work and it's not a big deal to maintain it (probably). If you don't agree with me - that's fine, we will then just merge it. It's generally a very small thing and we should probably use our time for discussions on other things :)

PR #835, opened 2018-06-23
jedi#setup_python_imports: better SyntaxError handling

blueyed:
Uses `repr` for SyntaxErrors.

Related: https://github.com/neovim/python-client/pull/221


davidhalter:
Makes sense!

PR #840, opened 2018-07-02
init: improve handling of import errors

blueyed:
See the individual commits.

TODO:
- [ ] squash before merging

### DIFF
--- a/pythonx/jedi_vim.py
+++ b/pythonx/jedi_vim.py
@@ -112,12 +112,7 @@ try:
     import jedi
 except ImportError:
     jedi = None
-    _, exc, tb = sys.exc_info()
-    while tb.tb_next:
-        tb = tb.tb_next
-    f_code = tb.tb_frame.f_code
-    exc_location = '%s:%d' % (f_code.co_filename, f_code.co_firstlineno)
-    jedi_import_error = '%s (in %s)' % (exc, exc_location)
+    jedi_import_error = sys.exc_info()
 else:
     try:
         version = jedi.__version__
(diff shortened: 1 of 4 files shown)


pythonx/jedi_vim.py
```diff
@@ -110,10 +110,14 @@ def echo_highlight(msg):
...
-except ImportError as e:
-    no_jedi_warning(str(e))
+except ImportError:
     jedi = None
-    jedi_import_error = str(e)
+    _, exc, tb = sys.exc_info()
+    while tb.tb_next:
+        tb = tb.tb_next
+    f_code = tb.tb_frame.f_code
+    exc_location = '%s:%d' % (f_code.co_filename, f_code.co_firstlineno)
```
davidhalter: Why don't we just return the whole traceback with `import traceback`? Wouldn't that be nicer and less code?

blueyed: Good point, but not trivial.  I will improve this.

blueyed:
Also uses PythonToVimStr now for saner string handling.

pythonx/jedi_vim_debug.py
```diff
@@ -1,57 +1,78 @@
...
+    echo(' - global site module: `{0}`'.format(__import__('site').__file__))
 
     try:
         import jedi_vim
-    except Exception as e:
-        echo("printf('ERROR: jedi_vim is not available: %s: %s', "
-             "{0!r}, {1!r})".format(e.__class__.__name__, str(e)))
+    except Exception:
+        echo_error('ERROR: jedi_vim is not available: {0}'.format(
+            format_exc_info(sys.exc_info())))
```
davidhalter: We can probably just remove sys.exc_info here, right?

blueyed: Yeah, I've added the default handling afterwards apparently.

blueyed: Fixed.

PR #847, opened 2018-07-15
qa: add flake8 to check make target

blueyed:


blueyed:
Thanks for merging so quickly.
I've meant to add this to it: https://github.com/davidhalter/jedi-vim/pull/848

PR #848, opened 2018-07-15
Makefile: use a single venv

blueyed:


davidhalter:
That is of course also fine :)

PR #849, opened 2018-07-15
show_call_signatures: skip empty params

blueyed:
Not sure if sigs for `Exception()` can be improved, but this patch handles this case in general.

FWIW, ipython displays the following for `Exception?`:
```
Init signature: Exception(self, /, *args, **kwargs)
Docstring:      Common base class for all non-exit exceptions.
Type:           type
```

### DIFF
--- a/pythonx/jedi_vim.py
+++ b/pythonx/jedi_vim.py
@@ -491,6 +491,10 @@ def show_call_signatures(signatures=()):
         # Descriptions are usually looking like `param name`, remove the param.
         params = [p.description.replace('\n', '').replace('param ', '', 1)
                   for p in signature.params]
+        if params == ['']:
+            # This is the case with `Exception(` for example.
+            # It makes no sense to display `()` then.
+            return
         try:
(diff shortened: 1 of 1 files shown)


davidhalter:
I'm not sure I like it. I just checked it in PyCharm (my co-worker uses it) and they do it in a totally different way, they just skip over both parentheses in case it does not have any parameters.

We don't really have a good way of showing the user that the function does not have any possible parameters, it will just look like Jedi doesn't know.

blueyed:
Oh, I see - so `()` could be considered to mean "no args", yes.

But for `Exception` it appears to be wrong, no?

davidhalter:
Yes for Exception it's just plain wrong, but I guess that's a jedi issue. You can also open a ticket there. I think it will probably be gone once we have proper support for typeshed.

blueyed:
Ok, closing this issue here.
Created https://github.com/davidhalter/jedi/issues/1182 to track it in jedi.

PR #850, opened 2018-07-15
Skip duplicate call signatures

blueyed:
I am seeing `p` twice for os.path.dirname, which seems to come from
Lib/posixpath.py and Lib/ntpath.py, as can be seen with `os.path.join`,
where I still see two with this patch:

```
            (path, *paths)
import os   (a, *p)
os.path.join()
```

btw: I had a hard time (i.e. failed) to get the filename information for the underlying definitions.
I think it might be useful to add additional information to the call signature in case there are multiple signatures and they are coming from different files/modules maybe?

### DIFF
--- a/pythonx/jedi_vim.py
+++ b/pythonx/jedi_vim.py
@@ -475,6 +475,7 @@ def show_call_signatures(signatures=()):
     if int(vim_eval("g:jedi#show_call_signatures")) == 2:
         return cmdline_call_signatures(signatures)
 
+    seen_sigs = []
     for i, signature in enumerate(signatures):
         line, column = signature.bracket_start
         # signatures are listed above each other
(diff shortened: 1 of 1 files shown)


davidhalter:
I think this is good for now.

> btw: I had a hard time (i.e. failed) to get the filename information for the underlying definitions.

I think there should be a `module_path` attribute on definitions?!

> I think it might be useful to add additional information to the call signature in case there are multiple signatures and they are coming from different files/modules maybe?

As I said in another place, there is going to be a param refactoring, but it's going to take some time. I'm also not sure how to solve that well. 

The problem is essentially the same for completions and it's not solved there, either. There we really just ignore completions if there are multiple ones with the same name. It's pretty hard to "merge" two definitions. I'm not sure how that would look in an API format.

PR #851, opened 2018-07-16
[RFC] Revisit usage highlighting

blueyed:
Pretty messy still (at least from the code), but some benefits:

- keeps definitions around until some text is changed (or the quickfix window is closed)
- keeps the qf window open on winleave
- uses `:cwindow` and no `:cclose` for better qf behavior in general, also gets rid of redraw! etc
- adds `p` for preview, although I find (mapped) `:cnext` etc better
- stores the actual matchids with the window

TODO:

- [x] text-properties instead of `matchaddpos()` (limited to current window) for Vim?

### DIFF
--- a/autoload/jedi.vim
+++ b/autoload/jedi.vim
@@ -471,10 +471,14 @@ function! jedi#add_goto_window(for_usages, len, ...) abort
 
     " Using :cwindow allows to stay in the current window in case it is opened
     " already.
+    let win_count = winnr('$')
     execute 'belowright cwindow '.height
+    let qfwin_was_opened = winnr('$') > win_count
 
-    if &filetype ==# 'qf'
-        " Setup quickfix window (initially, when opened via ":cwindow" above).
+    if qfwin_was_opened
+        if &filetype !=# 'qf'
+            echoerr 'jedi-vim: unexpected ft with current window, please report!'
+        endif
         if g:jedi#use_tabs_not_buffers == 1
             noremap <buffer> <CR> :call jedi#goto_window_on_enter()<CR>
         endif
--- a/autoload/jedi.vim
+++ b/autoload/jedi.vim
@@ -489,9 +489,6 @@ function! jedi#add_goto_window(for_usages, len) abort
                 autocmd WinLeave <buffer> q  " automatically leave, if an option is chosen
             endif
         augroup END
-
-        " Preview
-        " exe "nnoremap <buffer> p \<CR>:PythonJedi jedi_vim.highlight_usages_for_vim_win()\<CR>\<C-w>p"
     elseif a:for_usages && !s:supports_buffer_usages
         " Init current window.
         call jedi#_show_usages_in_win()
--- a/autoload/jedi.vim
+++ b/autoload/jedi.vim
@@ -461,8 +461,7 @@ endfunction
 " helper functions
 " ------------------------------------------------------------------------
 
-" a:mode: 'goto' or 'usages'
-function! jedi#add_goto_window(mode, len, ...) abort
+function! jedi#add_goto_window(for_usages, len, ...) abort
     let select_entry = a:0 ? a:1 : 0
     let height = min([a:len, g:jedi#quickfix_window_height])
 
@@ -477,10 +476,10 @@ function! jedi#add_goto_window(mode, len, ...) abort
         endif
 
         augroup jedi_goto_window
-            if a:mode ==# 'goto'
-                autocmd WinLeave <buffer> q  " automatically leave, if an option is chosen
-            else
+            if a:for_usages
                 autocmd BufWinLeave <buffer> call jedi#clear_usages()
+            else
+                autocmd WinLeave <buffer> q  " automatically leave, if an option is chosen
             endif
         augroup END
 
@@ -490,12 +489,12 @@ function! jedi#add_goto_window(mode, len, ...) abort
         if select_entry > 1
             exe select_entry.'cc'
         endif
-    elseif !has('nvim') && a:mode ==# 'usages'
+    elseif !has('nvim') && a:for_usages
         " Init current window.
         call jedi#show_usages_in_win()
     endif
 
-    if a:mode ==# 'usages'
+    if a:for_usages
         if has('nvim')
             augroup jedi_usages
               autocmd! BufWinEnter * call s:usages_for_buffer_nvim()
--- a/autoload/jedi.vim
+++ b/autoload/jedi.vim
@@ -301,9 +301,10 @@ function! jedi#usages() abort
     PythonJedi jedi_vim.usages()
 endfunction
 
+if !s:supports_buffer_usages
 " Hide usages in the current window.
 " Only handles the current window due to matchdelete() restrictions.
-function! jedi#hide_usages_in_win() abort
+function! jedi#_hide_usages_in_win() abort
     let winnr = winnr()
     let matchids = getwinvar(winnr, '_jedi_usages_vim_matchids', [])
 
@@ -319,8 +320,8 @@ function! jedi#hide_usages_in_win() abort
     unlet! b:_jedi_usages_needs_clear
 endfunction
 
-" Show usages for current window (Vim only).
-function! jedi#show_usages_in_win() abort
+" Show usages for current window (Vim without textprops only).
+function! jedi#_show_usages_in_win() abort
     PythonJedi jedi_vim.highlight_usages_for_vim_win()
 
     if !exists('#jedi_usages#TextChanged#<buffer>')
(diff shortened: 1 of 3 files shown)


autoload/jedi.vim
```diff
@@ -367,19 +395,47 @@ endfunction
...
+    " set lazyredraw
+    " try
+    "     execute 'belowright cwindow '.height
+    " finally
+    "     let &lazyredraw = save_lazyredraw
+    " endtry
+
+    " Using :cwindow allows to stay in the current window in case it is opened
+    " already.
+    execute 'belowright cwindow '.height
```
blueyed: @davidhalter 
Please comment on the code I've commented and are about to remove.

blueyed:
JFI: Vim v8.1.0218 allows to set matches for non-current windows, too.
Without this patch we could also still iterate over all windows.
Likely in a new PR then, just noticed this.

autoload/jedi.vim
```diff
@@ -288,16 +288,44 @@ function! jedi#goto_definitions() abort
...
+function! jedi#show_usages_in_win() abort
+    PythonJedi jedi_vim.highlight_usages_for_win()
+    if !exists('#jedi_usages#TextChanged')
+        augroup jedi_usages
+        " Unset highlights on any changes to this buffer.
+        autocmd TextChanged <buffer> call jedi#clear_usages()
+        " Hide usages when the buffer is removed from the window, or when
+        " entering insert mode (but keep them for later).
+        autocmd BufWinLeave,InsertEnter <buffer> call jedi#hide_usages_in_win()
+        augroup END
```
davidhalter: Can you indent everything in the augroup? I think it's way more readable the way you did it above.

autoload/jedi.vim
```diff
@@ -288,16 +288,44 @@ function! jedi#goto_definitions() abort
...
+    endif
+endfunction
+
+" Remove/unset global usages.
+function! jedi#clear_usages() abort
+    call jedi#hide_usages_in_win()
+    PythonJedi jedi_vim._current_highlights = None
+
```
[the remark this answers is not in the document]
davidhalter: And also _current_highlights is always a tuple in other places, shouldn't it also be a tuple here?

autoload/jedi.vim
```diff
@@ -385,19 +413,47 @@ endfunction
...
+    execute 'belowright cwindow '.height
+
     if g:jedi#use_tabs_not_buffers == 1
         noremap <buffer> <CR> :call jedi#goto_window_on_enter()<CR>
     endif
+
+    function! s:on_closed_qf() abort
+      " Remove global usages when the quickfix buffer is removed from the
+      " window, e.g. with :cclose.
+      PythonJedi jedi_vim.highlight_usages(None)
```
davidhalter: I think if you do this, I get a lot of `jedi-vim: called highlight_usages_for_win without highlights!`

Note that I use `g:jedi#use_tabs_not_buffers == 1`.

I'm not sure why this happens. However since it generally doesn't really clean up usages for me (neither with TextChanged, nor InsertEnter), I think there's something fundamentally wrong with the cleanup, but I don't really understand what it is.

blueyed:
Thanks for trying this and providing feedback.
I have been piled up some local WIP, which I am pushing now.
Not really tested with Vim, but it has better support for Neovim, which provides an API to set buffer highlights..!

autoload/jedi.vim
blueyed: This should be improved/fix, and/or needs a revisit again.

autoload/jedi.vim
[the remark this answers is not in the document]
blueyed: Adjusted.

blueyed:
Rebased.  /cc @weilbith

blueyed:
Another review would be appreciated.
(Not sure about how much I've addressed https://github.com/davidhalter/jedi-vim/pull/851#pullrequestreview-160843240 really, but just came back to it since usages highlighting in pretty broken currently when used across buffers, where it keeps highlight in the window from the previous buffer.)

pythonx/jedi_vim.py
```diff
@@ -370,9 +370,11 @@ def annotate_description(d):
     return '[%s] %s' % (typ, code)
 
 
-def show_goto_multi_results(definitions):
+def show_goto_multi_results(definitions, mode):
```
davidhalter: Can you maybe add a quick comment here why the different modes are needed? I'm not 100% sure why there is a different behavior for usages/goto. It might make a lot of sense, but when reading the code I wasn't sure why you implemented it that way.

autoload/jedi.vim
davidhalter: I don't really care about the code above. As long as it works - which it does it's fine. Also the reduction of complicated code is always a good thing.

autoload/jedi.vim
```diff
@@ -392,20 +461,61 @@ endfunction
...
-      au WinLeave <buffer> q  " automatically leave, if an option is chosen
-    augroup END
-    redraw!
+
+    " Using :cwindow allows to stay in the current window in case it is opened
+    " already.
+    execute 'belowright cwindow '.height
+
+    if &filetype ==# 'qf'
+        " Setup quickfix window, but only if it was opened / initially.
```
davidhalter: Can you comment a bit more on when it is opened and when not? I didn't really understand that. Also this if probably means that the focus is now in the quickfix window is selected (which it is only in goto mode?). I'm just not really sure when this triggers.

autoload/jedi.vim
```diff
@@ -392,20 +461,61 @@ endfunction
 " helper functions
 " ------------------------------------------------------------------------
 
-function! jedi#add_goto_window(len) abort
-    set lazyredraw
-    cclose
+" a:mode: 'goto' or 'usages'
```
davidhalter: Consider making this a boolean. There's quite a bit of confusion already around the different goto variants (stubs, assignments, definitions, etc.). `is_usages` would probably be way clearer.

blueyed: Good point. Used `for_usages`.

autoload/jedi.vim
blueyed: Changed the comment to:

> Setup quickfix window (initially, when opened via ":cwindow" above).

Clear?

autoload/jedi.vim
```diff
@@ -392,20 +461,61 @@ endfunction
...
+        augroup jedi_goto_window
+            if a:mode ==# 'goto'
+                autocmd WinLeave <buffer> q  " automatically leave, if an option is chosen
+            else
+                autocmd BufWinLeave <buffer> call jedi#clear_usages()
+            endif
+        augroup END
+
+        " Preview
+        " exe "nnoremap <buffer> p \<CR>:PythonJedi jedi_vim.highlight_usages_for_vim_win()\<CR>\<C-w>p"
```
blueyed: I guess this was meant for Vim to being able to highlight (from the quickfix window), since unlike Neovim highlights `matchaddpos()` only works for the current window.
Useful?

By now Vim has text-properties, which could/should be used probably instead of `matchaddpos()`.

pythonx/jedi_vim.py
blueyed: It gets passed through to `jedi#add_goto_window`, where it triggers different behavior.

blueyed:
> The location list (e.g. `:lwindow` instead of `:cwindow`) would probably be more appropriate.

Not really sure, since it often has global effects - but usually is local to the window then.
Would involve changing a settings name then maybe also though (`quickfix_window_height`).

The change for this is not big I think - and could even be made configurable then.

pythonx/jedi_vim.py
```diff
@@ -134,6 +134,58 @@ def echo_highlight(msg):
...
+    @classmethod
+    def call(cls, func, *args):
+        try:
+            f = cls._func_cache[func]
+        except KeyError:
+            if IS_NVIM:
+                f = cls._func_cache[func] = getattr(vim.funcs, func)
+            else:
+                f = cls._func_cache[func] = vim.Function(func)
+        return f(*args)
```
blueyed: This should / needs to be checked to how it behaves to the current method of going through `jedi#_vim_exceptions` / `_catch_exception`.
Would be necessary to get the "throwpoint" it seems.

autoload/jedi.vim
davidhalter: I guess I don't understand when this is not the case, because cwindow gets executed. While I know that cwindow can also close the window, I don't understand when that would trigger.

autoload/jedi.vim
davidhalter: Was this in there before? IMO it's not necessary for now.

pythonx/jedi_vim.py
davidhalter: Can we use for_usages here as well? That would be way clearer.

blueyed: There are different modes here after all (had that first, and then reverted), and the string gets used in the qf title.

autoload/jedi.vim
blueyed: The window might be open already.
Then cwindow does nothing.
Happens when using usages again while it's open, and if there's a window already (I guess).
I.e. that checks if we've moved to the window really.

pythonx/jedi_vim.py
davidhalter: Do you expect me to comment on this? Because I don't really understand it :) Which is fine of course, just asking.

pythonx/jedi_vim.py
davidhalter: I don't understand. It looks like it is only passed to `jedi#add_goto_window`?

pythonx/jedi_vim.py
blueyed: I've wondered if it is better/worse than the previous method, but then tested it.  I think it is good for now.

pythonx/jedi_vim.py
blueyed: `show_goto_multi_results` is also called for `goto`, which has different modes.
It uses `mode` the for the qf title (prefix).

blueyed:
I think this is ready by now.
Might give it a few more days for test-driving.
(should be squash-merged!)

davidhalter:
Feel free to merge whenever. It looks good to me. Thanks a lot!

blueyed:
\o/

PR #856, opened 2018-07-25
test/signatures.vim: remove trailing spaces

blueyed:


blueyed:
Working on https://github.com/davidhalter/jedi-vim/pull/652 to fix the tests for it.

davidhalter:
This is fine. I'm not a big fan of it, but I also don't mind the change.

blueyed:
We still have to find a fan of the current test setup.. :D

davidhalter:
:D

PR #858, opened 2018-07-27
Improve JediDebugInfo for envs

blueyed:
This is taken out of https://github.com/davidhalter/jedi-vim/pull/836.

### DIFF
--- a/pythonx/jedi_vim.py
+++ b/pythonx/jedi_vim.py
@@ -195,7 +195,7 @@ def get_environment(use_cache=True):
 
 def get_known_environments():
     """Get known Jedi environments."""
-    envs = list(jedi.api.environment.find_virtualenvs(paths=[], safe=False))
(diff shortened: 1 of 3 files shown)


pythonx/jedi_vim.py
```diff
@@ -193,6 +193,13 @@ def get_environment(use_cache=True):
     return environment
 
 
+def get_known_environments():
+    """Get known Jedi environments."""
+    envs = list(jedi.api.environment.find_virtualenvs(paths=[], safe=False))
```
davidhalter: Why overwrite safe? It's not like people are really using that (nor will they probably ever in jedi-vim) and also why paths?

blueyed: `paths` is used in #836 - thought it would be easier to merge then probably by keping it here.
´safe=False` is used because the user configures `paths` in #836.
I will change it here to not use any args.

davidhalter:
Thanks!

PR #860, opened 2018-07-29
tests: move/rename/cleanup vspec based tests

blueyed:
The main motivation is to being able to use normal/other pytest based
tests.

Therefore this moves everything from conftest into test_integration
itself.


davidhalter:
You can merge this after rebasing. I'm definitely fine with this!

blueyed:
I would like to get this in after #863 - then codecov would assure us that we're not dropping coverage due to this.

davidhalter:
Tests are failing and since you wanted to see the coverage changes, you might want to push again anyway :)

PR #863, opened 2018-07-29
Report coverage

blueyed:
TODO:

- [x] requires #864 
- [x] rename flags to py27/py36

Includes/rebased on #864 for now.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,6 @@
+[run]
+branch = 1
+source = .
+
+[report]
+include = pythonx/jedi_*.py,test/*
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,6 +1,7 @@
 [run]
 branch = 1
 source = .
+include = pythonx/jedi_*.py,test/*
 
 [report]
 include = pythonx/jedi_*.py,test/*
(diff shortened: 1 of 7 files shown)


blueyed:
For some reason py2 is not working: https://travis-ci.org/davidhalter/jedi-vim/jobs/409612945

`:checkhealth` reports it to be ok, but during tests `has('python')` appears to be false.

blueyed:
Found it:

> provider/pythonx: Could not load Python 2:^@          /home/travis/virtualenv/python2.7.14/bin/python2 is Python Coverage.py warning: --include is ignored because --source is set (include-ignored)^@          2.7 and cannot provide Python 2.

A Coverage.py warning apparently causes pytest to not detect py2 correctly.. :/

blueyed:
20.9% coverage.. \o/

Needs some cleanup, but should be good then already (let's get the branches/PRs it depends on though in anyway).

Next step would be covimerage to track coverage for the Vimscript part.

blueyed:
Should be good, but I want to cleanup/review it myself once more.

blueyed:
I like it! ;)

Should be squash-merged (I can do it).

davidhalter:
I'll just put you in charge for this. :) I like that we have coverage and don't really care how it's exactly done!

blueyed:
Yeah! \o/

PR #864, opened 2018-07-29
Test with Neovim also

blueyed:
This will allow us to get coverage via its Python subprocess then.

Includes #862 for setting `VSPEC_VIM`.


davidhalter:
@blueyed Can you rebase. Sorry I think I squashed and this messed it up.

PR #867, opened 2018-08-01
Fix displaying of error information with shortmess+=F

blueyed:
When being initialized via ftplugin `set shortmess+=F` (default in
Neovim by now) causes the additional information (the full traceback) to
be silenced.

This patch uses `:unsilent` to always display it.

Ref: https://github.com/neovim/neovim/issues/8675


davidhalter:
Makes sense! Thanks.

PR #872, opened 2018-08-13
jedi_vim.get_environment: cache with $VIRTUAL_ENV

blueyed:
Note that you have to use e.g. `:py3 os.environ['VIRTUAL_ENV'] = 'foo'`
to update this, `:let $VIRTUAL_ENV = 'foo'` is not enough (both in
Neovim and Vim).

Ref: https://github.com/davidhalter/jedi/issues/1200

### DIFF
--- a/pythonx/jedi_vim.py
+++ b/pythonx/jedi_vim.py
@@ -164,12 +164,14 @@ def get_environment(use_cache=True):
     global current_environment
 
     vim_force_python_version = vim_eval("g:jedi#force_py_version")
-    if use_cache and vim_force_python_version == current_environment[0]:
+    virtual_env = os.environ.get('VIRTUAL_ENV')
+    cache_key = (vim_force_python_version, virtual_env)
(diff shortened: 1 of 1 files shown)


blueyed:
Re coverage: it appears to not mark anything covered after e.g. `vim.eval` is used.. :/
Filed https://github.com/neovim/python-client/issues/354, but am not sure if this is a neovim-python-client problem really.

davidhalter:
Hmm, I haven't looked at it in detail, but wouldn't the proposed Jedi change of yours be good enough?

blueyed:
I think this still makes sense even with the improvement in Jedi, right?
https://github.com/davidhalter/jedi/commit/862f61182941a451c8e158091a328d63cb4ef43b

davidhalter:
> I think this still makes sense even with the improvement in Jedi, right?

I agree that this still made sense at the time. However with the latest changes it's not necessary anymore, because we do caching.

I'm closing, but feel free to reopen if you think that this is still relevant.

PR #901, opened 2019-01-24
Update parso to v0.3.2

### LINKED ISSUE #48: No completion options for tarfile
For some reason `jedi#complete` gets confused by

``` python
import tarfile

tarfile.
```

and doesn't show any completion options.

This is Python 2.7 on Ubuntu 12.10.

### LINKED ISSUE #51: Rebind autocompletion?
Is there a way to easily rebind the autocompletion feature? For example, from control + space to shift + space? It might also be useful to put this information into the readme somewhere for VIM newbs like me :)

blueyed:
Update submodule pythonx/parso 930ec08...a2b153e

    * pythonx/parso 930ec08...a2b153e (91):
      > Upgrade the Changelog
      > Escape a backslash properly
      > Bump the parso version to 0.3.2
      > f-string parts are also PythonLeaf instances
      > Fix get_start_pos_of_prefix
      > Remove dead code
      > Refactor f-string support
      > Fix: End detection for strings was mostly wrong, fixes #51
      > Fix f-string escapes, fixes #48
      > Tokenizer: Simplify end of string regexes
      > Add a comment from the Python3.7 code base
      > Fix diff parser: The previous fix was a bit off
      > Change fuzzer: Add ways to not always use correct parse input
      > Fix diff parser: issue with opening brackets
      > Async doesn't work in 3.4
      > Skip some tests for Python2.6 and Python3.3
      > Fix diff parser: DEDENT as error leaves should also be ignored and reparsed
      > Fix diff parser: error dedents in between nodes should be ignored for now when copying
      > Fix diff parser: Don't copy error nodes/leaves in the beginning, leads to strange issues in some cases
      > Diff fuzzer: Create a check to see if the errors make sense.
      > Fix diff parser: Don't copy DEDENT tokens at the beginning
      > Fix diff parser: Skip last leaves for last line offset leaves
      > Fix indentation error tokens
      > Fix diff parser: Avoid side effects for prefix
      > Fix tokenizer: backslashes sometimes led to newline token generation
      > Make lines easier copyable in the fuzzer
      > Fix tokenizer: Carriage returns after backslashes were not properly handled
      > Fix in diff parser: prefix calculation was wrong when copying nodes
      > Replace --print-diff with --print-code
      > Fix diff parser generation for empty files
      > Docopt should not be needed for tests
      > Fix tokenizer: Form feeds and multiline docstrings didn't work together
      > Fix diff parser: Need to calculate the prefix for the diff tokenizer better
      > Remove copied nodes stuff, to simplify some things
      > Fix newline endings and a few parser/copy counts
      > Feature: The diff parser fuzzer is now able to use random Python fragments
      > Fix: The Python 3.8 grammar did not include f-string support
      > Fix tokenizer: Closing parentheses in the wrong place should not lead to strange behavior
      > Fix end positions with error dedents
      > Fix diff parser tests for Python 2
      > Randomize the fuzzer a bit more with inserting characters
      > Simplify the regexes
      > Fix a regex clause that was totally wrong
      > Fix another tokenizer issue
      > Apply \r changes in syntax trees
      > Actually make \r usable
      > Fix split lines for Python code
      > Add \r as a valid linebreak for splitlines
      > Move split_lines tests around
      > Start generating really random strings with the fuzzer
      > Fix tokenizer: Dedents before error tokens are properly done, now.
      > Fix diff parser: Also check async with
      > Make the diff parser use a lot of different files by default
      > Fix diff parser: Forgot that with statments are also flows
      > Fix valid graph asserting for some dedents that are errors
      > Fix diff parser: If funcs are not copied, errors shouldn't either
      > Better documentation for the fuzz diff parser script
      > Rename a test
      > More verbose output for the diff fuzzer
      > Fix diff parser: error dedent issues
      > Fix diff parser: Avoid indentation issues
      > Use the right diff order in debug output
      > Fix diff parser: The prefix was wrong in some copy cases
      > Make fuzz parser compatible with Python 2
      > Make it possible to print all diffs in fuzzer
      > Introduce a redo flag 'only_last' to narrow down issues
      > Add docopt to testing dependencies
      > Rewrite the fuzz diff parser to cache errors (so we can re-run those)
      > Fix diff parser: Decorators were sometimes parsed without their functions
      > Fix tokenizer: Dedents should only happen after newlines
      > Usability for diff parser fuzzing
      > Fix for diff parser: Rewrite prefix logic and don't mutate prematurely
      > Fix diff parser: positioning of functions if decorators were removed
      > Ignore ERROR_DEDENT in graph validation
      > Add a small diff parser fuzzer
      > Fix diff parser: get_last_line was sometimes wrong
      > Fix error reporting order for diff issues
      > Clean up the test diff parser file
      > Avoid nasty side effects in creation of Node
      > Make parso errors even more informative
      > Fix diff parser get_last_line calculation
      > Fix graph asserting for error indents
      > Better checks for checking valid graphs
      > Create better ways for debugging the diff parser
      > Fix diff-parser: Copying parts of if else should not lead to the whole thing being copied
      > Reenable diff parser parser counting in all tests
      > Fix: Make the NodesStack to a NodesTree
      > Fix for diff parser : Make sure that start_pos are growing always The problem was that functions/classes were sometimes not well positioned. Now all diff tests are ensuring that leaves always grow.
      > Add a check to see if leaves have the right start positions
      > Add parso version to an exception
      > MNT: add grammar for python 3.8


davidhalter:
Thanks!

PR #906, opened 2019-02-03
Open all folds before doing a jedi rename

### LINKED ISSUE #789: Problem when rename variable inside folds
### Issue

Using Python, when I rename a variable inside a method and this same method has a folded section, if the word to be renamed also exists in this fold, the entire fold will be replaced by the new variable name.

### Steps to reproduce

1. write a function
```
old_name1 = 'aaaa'
def test():
    old_name2 = 'bbbb'
    for i in range(4):
        for j in range(4)
            a = 1
            print(old_name1[i])
            print(old_name2[j])
            b = 2

```
2. :setlocal foldmethod=indent

3. fold at the second `for`, rename old_name1 to new_name1

4. all the folded block will be renamed as new_name1

stevanmilic:
If folds are present before the rename they will
be overwritten if folds aren't open, because
`vim.current.window.cursor` doesn't jump inside
the folds.

Fixes #789


pythonx/jedi_vim.py
```diff
@@ -716,6 +716,10 @@ def do_rename(replace, orig=None):
         # Save view.
         saved_view = vim_eval('string(winsaveview())')
 
+        # Open all folds in the window, so that replacing of original
+        # word doesn't overwrite the folded section.
+        vim_command('normal! zR')
+
         # Replace original word.
         vim.current.window.cursor = (r.line, r.column)
         vim_command('normal! c{0:d}l{1}'.format(len(orig), replace))
```
blueyed: I think we should use another API to change the word instead, likely the `setline()` function.
This avoids (re)storing the view also.

blueyed: You could just set `vim.current.buffer[r.line - 1]` to the resulting string (where the replacement was done in/via Python).

stevanmilic: oh nice, thanks! should be solved now

blueyed:
Great!

CI failure is unrelated, working on fixing it in https://github.com/davidhalter/jedi-vim/pull/908.

blueyed:
Please rebase it on master.

stevanmilic:
done

PR #915, opened 2019-03-09
With typeshed coming to Jedi, builtin modules should be displayed

davidhalter:
(If there's a module_path, which most will have)

This change is still backwards compatible and can be used with older Jedi
versions (<0.14.0).

We can merge this now or later. IMO it's good to merge now, because it won't really change the behavior AFAIK.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -82,9 +82,8 @@ Manual installation
 You might want to use `pathogen <https://github.com/tpope/vim-pathogen>`_ or
 `Vundle <https://github.com/gmarik/vundle>`_ to install jedi-vim.
 
-The first thing you need after that is an up-to-date version of Jedi. You can
-either install it via ``pip install jedi`` or with
-``git submodule update --init`` in your jedi-vim repository.
+The first thing you need after that is an up-to-date version of Jedi. Install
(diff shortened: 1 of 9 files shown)


blueyed:
I trust you on this, but tests need to be fixed.

davidhalter:
I think the major change is that there's now a `<leader>s` command that goes to stubs.

Please review and maybe test. We can then merge. Pretty sure it works.

davidhalter:
BTW: I'm a bit unsure about the test failures. If you have any idea, I'm interested.

davidhalter:
Tests are passing now. I feel like I finally fixed it.

blueyed:
Awesome!
Checked this out locally now and will test-drive it.

davidhalter:
Not sure what you intend to do now. Are you merging once you tested it for a bit?

blueyed:
Feel free to merge it already (the approval is based on reviewing).
I've not tested it much yet, but changes to jedi-vim are minor, and I've used updated jedi/parso before, so that's fine.

davidhalter:
Merging, worst case we revert.

PR #926, opened 2019-04-10
add g:jedi#virtualenv_path to specify virtualenv

theoldmoon0602:
Hello.

I added `g:jedi#vritualenv_path` to specify the virtual environment like `$VIRTUAL_ENV`, but this will work even after Jedi is started.

Would you please merge if you think this feature is good.

The following script is my usage.

```vim
" set the pipenv's environment for completion when open .py file
function! s:setPipenvPath()
  let pipenv_dir = expand('%:p:h')
  if !filereadable(l:pipenv_dir . '/Pipfile')
    return
  endif

  let venv_path = trim(system(printf("sh -c 'cd %s; pipenv --venv'", pipenv_dir)))
  let g:jedi#virtualenv_path = venv_path
endfunction

autocmd FileType python :call s:addPipenvPath()
```

### DIFF
--- a/autoload/jedi.vim
+++ b/autoload/jedi.vim
@@ -34,6 +34,7 @@ let s:default_settings = {
     \ 'popup_select_first': 1,
     \ 'quickfix_window_height': 10,
     \ 'force_py_version': "'auto'",
+    \ 'virtualenv_path': "''",
     \ 'smart_auto_mappings': 0,
     \ 'use_tag_stack': 1
 \ }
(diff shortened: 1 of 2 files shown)


blueyed:
I think we should finish https://github.com/davidhalter/jedi-vim/pull/836 instead.

theoldmoon0602:
@blueyed I see. Thank you :)

blueyed:
Thanks for your PR in the first place anyway, of course!

Please give #836 a try/review.

PR #927, opened 2019-04-19
Fix `:JediClearCache!`

blueyed:
Parso's cache can be cleaned from Jedi now.


davidhalter:
I'm fine with this change. However I feel like if we actually want to have such a command properly we should remove `jedi.settings.cache_directory` and clear parso's in memory cache. However parso doesn't really have a stable API to do that (yet).

The time caches are pretty irrelevant IMO, it's probably not what people care about.

I'll still merge, because it's better than this just breaking.

blueyed:
Yes, I've noticed that it wasn't enought and a Vim restart was necessary.

Issue #213, opened 2013-11-13
Add an option to disable key bindings


Julian:
Hey,

Would you be willing to add an option to disable binding keys? (Apologies if it exists already, I only see `g:jedi#auto_initialization = 0` but that disables more than just mappings).

I know you can remap each key to make it use a different key, but that means that users have to go look up each map and pick some other key, even if they don't care about it (i.e., I just want Jedi to not quash my mappings, I don't really care about all the functionality it's mapping).

Another alternative (or something to consider) based on the above would be to not map keys if they're already mapped to something else.

sloria:
Sorry to post on an old issue, but I still think this would be a useful feature. I, too, was looking for an option to disable setting of the default keymappings (without turning off other initialization steps, e.g. omnifunc).

That said, I respect that you want to keep jedi-vim's configuration/API small.

blueyed:
We might at least use `:map-<unique>`, which would display an error if a mapping is used already.
Or to be more friendly: check this ourselves and provide helpful instructions.

I also think that there are a lot of settings already.
Best practice would be to use `<Plug>` mappings, instead of settings for the left-hand-side of the map.
We could then check if a map to something is provided already (which means the user has defined it manually).

But it would need to handle the old settings for B/C and therefore it's probably not worth it.

davidhalter:
I personally still think that the current way is ok. I would be willing to use `unique`, if that helps. But I don't really want to add new settings.

Julian:
`unique` is the wrong direction (would be worse) IMO. It's my mapping I want, `jedi` shouldn't cause errors on startup :).

Julian:
Just as another example of why the current solution isn't nice to users (well, really to me, I can't speak for anyone else), a recent commit broke my leader mappings once again, because `g:goto_command` was added, which I wasn't overriding.

dbrodie:
I agree with @Julian, I would like a feature like this as well, especially since I don't use jedi-vim as much by itself rather than integrate it into existing completion frameworks (such as deoplete, neocomplete) so that it is consistent with other languages I program in.

I think the best solution is to use the <Plug> mapping and have a setting to disable the mapping, or something equivalent.

blueyed:
You're welcome to provide a PR.

> I don't use jedi-vim as much by itself rather than integrate it into existing completion frameworks (such as deoplete, neocomplete) 

[...]

PR #943, opened 2019-08-21
Add option to disable default keybindings.

kadaliao:
#213

### DIFF
--- a/doc/jedi-vim.txt
+++ b/doc/jedi-vim.txt
@@ -203,6 +203,9 @@ example, to set the keybinding for starting omnicompletion to <C-N> instead of
 
     let g:jedi#completions_command = "<C-N>"
 
+Note: If you have |g:jedi#disable_default_mappings| set to 1, you can disable
+all the default keybindings.
+
 Note: If you have |g:jedi#auto_initialization| set to 0, you have to create
 a mapping yourself by calling a function: >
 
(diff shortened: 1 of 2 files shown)


blueyed:
Well, quoting @davidhalter:

> You can just set all the mappings to "" and nothing will happen to your settings anymore. Considering that it's just 6 settings (and you probably won't disable all of them because you want some of them!)

Maybe the doc with "auto_initialization" should just provide an example to unset all of them?
Then you can just add those 6 lines to your config, if you want to delete all of them.

kadaliao:
> Well, quoting @davidhalter:
> 
> > You can just set all the mappings to "" and nothing will happen to your settings anymore. Considering that it's just 6 settings (and you probably won't disable all of them because you want some of them!)
> 
> Maybe the doc with "auto_initialization" should just provide an example to unset all of them?
> Then you can just add those 6 lines to your config, if you want to delete all of them.

Oh yes. I started this because I wanted to change the default mappings to avoid conflict with other plugins. Just found out that I can directly change the mappings by setting the command string.

I guess the explicit disable option will be unnecessary. 😞

blueyed:
@kadaliao 
No worries, and thanks for the PR in the first place!

I've created https://github.com/davidhalter/jedi-vim/pull/944 as a follow-up.

Issue #953, opened 2019-09-26
jedi-vim autocompletion shows the same module twice instead of class after dot


iGr33k:
### Issue

Problem occurs with the following code:
`import struct`
`import pydbg`
`dbg = pydbg.pydbg()`
`dbg.bp_set()`

When i try to autocomplete `bp_set()` part it returns `Pattern not found`
However when i type `dbg = pydbg.pydbg.pydbg()` and then `bp_set` everything is Ok with jedi-vim but code obviously doesn't work.
### Steps to reproduce

my minimal.vimrc:
`set nocompatible`
`let script_dir = fnamemodify(expand('<sfile>'), ':h')`
`let &runtimepath .= ','.script_dir.','.script_dir.'/after'`

`" Put your config changes here.`
`execute pathogen#infect()`
`" let g:jedi#show_call_signatures=1`

`syntax on`
`filetype plugin indent on`

General steps have been covered in first section already.
pydbg library can be found [here](https://github.com/OpenRCE/pydbg)
### Output of “:verbose JediDebugInfo”

```
#### Jedi-vim debug information


##### jedi-vim version


 - jedi-vim git version: 'git' is not recognized as an internal or external command,
operable program or batch file.
 - jedi git submodule status: 'git' is not recognized as an internal or external command,
operable program or batch file.
 - parso git submodule status: 'git' is not recognized as an internal or external command,
operable program or batch file.


##### Global Python


Using Python version 2 to access Jedi.
 - global sys.executable: `C:\Program Files (x86)\Vim\vim81\gvim.exe`
 - global sys.version: `2.7.16 (v2.7.16:413a49145e, Mar  4 2019, 01:30:55) [MSC v.1500 32 bit (Intel)]`
 - global site module: `C:\Python27\Lib\site.pyc`

##### Jedi

 - path: `C:\Python27\lib\site-packages\jedi\__init__.pyc`
 - version: 0.15.1

##### Jedi environment: <Environment: 2.7.16 in C:\Python27>


 - executable: C:\Python27\python.exe
 - sys_path:
    - `C:\Windows\SYSTEM32\python27.zip`
    - `C:\Python27\DLLs`
    - `C:\Python27\lib`
    - `C:\Python27\lib\plat-win`
    - `C:\Python27\lib\lib-tk`
    - `C:\Python27`
    - `C:\Python27\lib\site-packages`

##### Known environments


 - <Environment: 2.7.16 in C:\Python27> (C:\Python27\python.exe)

[...]

PR #954, opened 2019-09-27
debug info: catch any errors with PythonJedi commands

blueyed:
Ref: https://github.com/davidhalter/jedi-vim/issues/953


davidhalter:
Thanks!

Issue #948, opened 2019-09-19
Some modules not found


scmartin:
### Issue

Trying to use jedi-vim with a python3 conda environment, I find that some modules are found but others are not, and result in "Omni completion (^O^N^P) Pattern not found. Some modules that are not working are
scipy
numpy
sympy
pandas

Examples of modules that are working
matplotlib
jsonschema
markupsafe

### Steps to reproduce

With the following minimal vim setup:
-   set nocompatible              " be iMproved, required
-   filetype off                  " required
-   let g:pymode_python = 'python3'
-   let &pythonthreedll = '/panfs/pfs.local/software/7/install/anaconda/4.7/lib/libpython3.7m.so.1.0'
-   color elflord
-   syntax on
-   filetype plugin on
-  set nu

Running ":python3 import numpy; print(np.arange(10))" returns the expected array.
" :python3 import sys; print(sys.path) " includes the path to my conda environment.

Not sure what else to try.

### Output of “:verbose JediDebugInfo”

 - jedi-vim git version: Unknown option: -C
usage: git [--version] [--help] [-c name=value]
           [--exec-path[=<path>]] [--html-path] [--man-path] [--info-path]
           [-p|--paginate|--no-pager] [--no-replace-objects] [--bare]
           [--git-dir=<path>] [--work-tree=<path>] [--namespace=<name>]
           <command> [<args>]
 - jedi git submodule status: Unknown option: -C
usage: git [--version] [--help] [-c name=value]
           [--exec-path[=<path>]] [--html-path] [--man-path] [--info-path]
           [-p|--paginate|--no-pager] [--no-replace-objects] [--bare]
           [--git-dir=<path>] [--work-tree=<path>] [--namespace=<name>]
           <command> [<args>]
 - parso git submodule status: Unknown option: -C
usage: git [--version] [--help] [-c name=value]
           [--exec-path[=<path>]] [--html-path] [--man-path] [--info-path]
           [-p|--paginate|--no-pager] [--no-replace-objects] [--bare]
           [--git-dir=<path>] [--work-tree=<path>] [--namespace=<name>]
           <command> [<args>]

##### Global Python

Using Python version 3 to access Jedi.
 - global sys.executable: `/home/sether35/.conda/envs/notebook/bin/python3`
 - global sys.version: `3.7.3 (default, Mar 27 2019, 22:15:47), [GCC 7.3.0]`
[...]

PR #958, opened 2019-10-19
completions: skip docstr without preview, handle exceptions

blueyed:
Closes https://github.com/davidhalter/jedi-vim/issues/948.

### DIFF
--- a/pythonx/jedi_vim.py
+++ b/pythonx/jedi_vim.py
@@ -305,16 +305,22 @@ def completions():
             completions = script.completions()
             signatures = script.call_signatures()
 
+            add_info = "preview" in vim.eval("&completeopt").split(",")
             out = []
             for c in completions:
                 d = dict(word=PythonToVimStr(c.name[:len(base)] + c.complete),
                          abbr=PythonToVimStr(c.name_with_symbols),
                          # stuff directly behind the completion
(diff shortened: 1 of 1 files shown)


pythonx/jedi_vim.py
```diff
@@ -305,16 +305,22 @@ def completions():
...
-                         info=PythonToVimStr(c.docstring()),  # docstr
                          icase=1,  # case insensitive
                          dup=1  # allow duplicates (maybe later remove this)
                          )
+                if add_info:
+                    try:
+                        d["info"] = PythonToVimStr(c.docstring())
+                    except Exception:
+                        print("jedi-vim: error with docstring for %r: %s" % (
+                            c, traceback.format_exc()))
```
davidhalter: Just wondering: Do we need that? Or is this already covered by the generic error handling and you just want to know a bit more about the error (essentially the added `c` information)?

Otherwise I think this is a good change. Not sure though how many people aren't using preview. How about you?

blueyed: I am not using preview, but it is enabled by default IIRC.
I've thought it is good to log/print errors (otherwise we might have never noticed that it fails with parts of numpy in the first place likely).  It will only show up with `:mess` though.
The decorator would only print it when not swallowed (what is done here now).

PR #966, opened 2019-12-04
Make sure call signatures are cleared before they are calculated again

davidhalter:


davidhalter:
@blueyed I'm merging this, since it show clear improvements. 

I have been testing quite a bit to avoid future things like https://github.com/davidhalter/jedi/issues/1399. This was one problem (of about 3, the other two being part of Jedi).

IMO #652 is still relevant. I just think that this change is very small and can therefore take priority. Also this bug is present in #652 as well.

blueyed:
@davidhalter 
Thanks!
I assume this is needed since Jedi picks up the text with the concealed signatures otherwise?
Do you have steps to reproduce this?  With #652 it should only be relevant when still using the conceal method there I assume?

pythonx/jedi_vim.py
```diff
@@ -765,9 +765,12 @@ def show_call_signatures(signatures=()):
     if int(vim_eval("has('conceal') && g:jedi#show_call_signatures")) == 0:
         return
 
+    # We need to clear the signatures before we calculate them again. The
+    # reason for this is that call signatures are unfortunately written to the
+    # buffer.
+    clear_call_signatures()
     if signatures == ():
         signatures = get_script().call_signatures()
-    clear_call_signatures()
```
blueyed: This should be needed only inside the `if` probably (strictly speaking)?
(it's ok to keep it like that for now though)

blueyed:
Related code for completions btw: https://github.com/davidhalter/jedi-vim/blob/12e97c7a04f8fd80afe0de0f1c9181f2c7bc53ca/pythonx/jedi_vim.py#L278-L283

davidhalter:
> With #652 it should only be relevant when still using the conceal method there I assume?

Yes. Exactly.

pythonx/jedi_vim.py
davidhalter: True. We could move it there.

This was just my poor man's fix to avoid any regressions. If you want to further improve this, feel free. I didn't put a lot of thought into it. It's a simple move that should not change anything (except fix the original issue).

PR #967, opened 2019-12-05
ci: Travis: slighty improve codecov-bash invocation

blueyed:


davidhalter:
Feel free to merge yourself. I cannot really judge this and if it's not working, I don't think it's that worrying :)

PR #968, opened 2019-12-05
Fix invalidating cache with `jedi#show_call_signatures`

blueyed:
Without this it would not redraw properly when moving left within `print(…)` - see the test.

### DIFF
--- a/autoload/jedi.vim
+++ b/autoload/jedi.vim
@@ -581,15 +581,16 @@ function! jedi#show_call_signatures() abort
 
     " Caching.  On the same line only.
     if line == s:show_call_signatures_last[0]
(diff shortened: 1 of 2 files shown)


blueyed:
Should be squash-merged (or just approved ;)).

blueyed:
@davidhalter 
The problem here is the `=` - moving across it makes Jedi return a new signature (with different selected argument), so we must invalidate the cache when the number of "="s changes left/right to the cursor also.
Does that make sense?  Did you mean something else?

davidhalter:
I understood that part. I just didn't really grasp the whole logic. Looked at it again, now I get it and I think it's all good.

PR #980, opened 2020-01-10
minor: rewrite/expand show_documentation

blueyed:
The documentation test fails with newer Jedi, this is for investigating it.

```
________________ test_integration[test/vspec/documentation.vim] ________________
test/vspec/documentation.vim failed:
"__doc__" [New File]
# ImportError(*args, name: Optional[str]=..., path: Optional[str]=...)
not ok 1 - documentation docstrings simple
# Expected stridx(content, "Import can't find module") > 0 at line 11
#       Actual value: -1
#     Expected value: 0
5 buffers wiped out
ok 2 - documentation docstrings no documentation
1..2
```

TODO:

- [x] failure on py27

      test/vspec/documentation.vim failed:
      not found in 'runtimepath': "ftdetect/*.vim"
      not ok 1 - documentation docstrings simple
      # Expected header[0] == "Docstring for builtins:class ImportError" at line 7
      #       Actual value: "Docstring for __builtin__:class ImportError"
      #     Expected value: "Docstring for builtins:class ImportError"
      5 buffers wiped out
      ok 2 - documentation docstrings no documentation
      1..2

### DIFF
--- a/pythonx/jedi_vim.py
+++ b/pythonx/jedi_vim.py
@@ -726,9 +726,17 @@ def show_documentation():
     if not definitions:
         echo_highlight('No documentation found for that.')
         vim.command('return')
-    else:
-        docs = ['Docstring for %s\n%s\n%s' % (d.desc_with_module, '=' * 40, d.docstring())
-                if d.docstring() else '|No Docstring for %s|' % d for d in definitions]
+        return
+
+    docs = []
+    for d in definitions:
+        doc = d.docstring()
+        if doc:
+            title = 'Docstring for %s' % d.desc_with_module
+            underline = '=' * len(title)
+            docs.append('%s\n%s\n%s' % (title, underline, doc))
+        else:
+            docs.append('|No Docstring for %s|' % d)
         text = ('\n' + '-' * 79 + '\n').join(docs)
         vim.command('let l:doc = %s' % repr(PythonToVimStr(text)))
         vim.command('let l:doc_lines = %s' % len(text.split('\n')))
--- a/test/vspec/documentation.vim
+++ b/test/vspec/documentation.vim
@@ -15,7 +15,16 @@ describe 'documentation docstrings'
         normal GK
         Expect bufname('%') == "__doc__"
         Expect &filetype == 'rst'
-        let content = join(getline(1,'$'), "\n")
+        let header = getline(1, 2)
+        PythonJedi vim.vars["is_py2"] = sys.version_info[0] == 2
+        if g:is_py2
+            Expect header[0] == "Docstring for __builtin__:class ImportError"
+            Expect header[1] == "==========================================="
+        else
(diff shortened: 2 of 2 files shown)


pythonx/jedi_vim.py
```diff
@@ -724,9 +724,17 @@ def show_documentation():
     if not definitions:
         echo_highlight('No documentation found for that.')
         vim.command('return')
-    else:
-        docs = ['Docstring for %s\n%s\n%s' % (d.desc_with_module, '=' * 40, d.docstring())
-                if d.docstring() else '|No Docstring for %s|' % d for d in definitions]
+        return
```
davidhalter: This is probably too much and should be removed :) (Probably the whole if/else)

blueyed: @davidhalter 
What do you mean exactly?
I'd rather not change it too much - I wanted to investigate only why Jedi master breaks it.

And it should be rewritten to use `script.help()` instead then better anyway.

test/vspec/documentation.vim
```diff
@@ -15,7 +15,16 @@ describe 'documentation docstrings'
...
+        PythonJedi vim.vars["is_py2"] = sys.version_info[0] == 2
+        if g:is_py2
+            Expect header[0] == "Docstring for __builtin__:class ImportError"
+            Expect header[1] == "==========================================="
+        else
+            Expect header[0] == "Docstring for builtins:class ImportError"
+            Expect header[1] == "========================================"
+        endif
+        let content = join(getline(3, '$'), "\n")
         Expect stridx(content, "Import can't find module") > 0
```
blueyed: @davidhalter with Jedi master "content" is "ImportError(*args, name: Optional[str]=..., path: Optional[str]=...)" here, i.e. the signature itself?!
git-bisected to https://github.com/davidhalter/jedi/commit/4161bfc7f28454db2a307cdcae463e41bec3103d

pythonx/jedi_vim.py
davidhalter: Oh, I realized that this is probably ok, I didn't really understand what was changed here.

test/vspec/documentation.vim
davidhalter: @blueyed For whatever reason, the documentation of some builtins doesn't seem to work anymore and the tests don't catch it. No idea why, I will investigate and fix it. I have noticed it a few days ago when documentation for open didn't show up. So really a Jedi bug.

blueyed: It works with the submodule's version, doesn't it?

blueyed: I.e. the test fails locally when updating Jedi to the above commit.
Going to merge this for now.

davidhalter: Make sense, thanks!

PR #990, opened 2020-01-29
Fix "goto" with multiple results

blueyed:
E.g. with `os.path`.

### DIFF
--- a/pythonx/jedi_vim.py
+++ b/pythonx/jedi_vim.py
@@ -434,6 +434,8 @@ def show_goto_multi_results(definitions, mode):
     """Create (or reuse) a quickfix list for multiple definitions."""
     global _current_definitions
 
+    for_usages = mode == "usages"
+
     lst = []
     (row, col) = vim.current.window.cursor
     current_idx = None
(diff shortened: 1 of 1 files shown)


davidhalter:
I'll try to test it locally.

blueyed:
> I'll try to test it locally.

:+1:
Works on `os.path`, where it should give you two results to choose from.

davidhalter:
It seems like goto is working again like I assumed.

The only thing I'm not happy about now is the `cc` for usages (which was in there before :)). I feel like this makes it harder to actually go to something (and most people like me don't know about `:ccl`, which is also pretty annoying to type). My workflow is usually that I use usages and then just go to one of them. Feel free to comment, maybe your workflow is better and I just don't know how to deal with it.

But: This can be merged, since it just improves the situation.

davidhalter:
BTW: My issue might just be the focus of the window. IMO selecting the current usage makes sense, but leaving the window doesn't.

What do you think?

blueyed:
> BTW: My issue might just be the focus of the window. IMO selecting the current usage makes sense, but leaving the quickfix window doesn't.
> 
> What do you think?

I can see what you mean.
My use is just seeing where it is used, but it makes sense to select it explicitly.  Will look into it later.

blueyed:
@davidhalter pushed a fixup.

davidhalter:
Makes sense, I guess. Thanks!

PR #991, opened 2020-01-29
Update parso/jedi (v0.6.0, v0.16.0+)

blueyed:

PR #993, opened 2020-02-02
tests: cover jedi_debug_info

blueyed:
TODO:

- [x] after https://github.com/davidhalter/jedi-vim/pull/991

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ install:
 script:
   - make --keep-going "$ENV" BUILD_VIRTUAL_ENV=$VIRTUAL_ENV
 
-after_success:
+after_script:
   - |
     if [ "$ENV" = "test_coverage" ]; then
       coverage xml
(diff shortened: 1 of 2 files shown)


blueyed:
51% at least.. ;)

Fails on py27 due to:

> Error when running display_debug_info: Vim(return):Exception TypeError: '__call__() takes exactly 1 argument (2 given)' in <jedi._compatibility.finalize object at 0x7f0f8b98e360> ignored

IIRC this might be fixed with the update (#991).

davidhalter:
BTW: The question is also how long we are going to support Python 2. I have deprecated it in Jedi and I will remove it after a while. I guess most VIM versions (including my old one) come with Python 3 support. Jedi can still deal with Python 2 (for a while), so people could just use `:let jedi#force_py_version=2.7` for the time being.

blueyed:
@davidhalter IMHO it makes sense to remove it gets into the way, i.e. when we want to add type annotations etc, but then this would be easier after also dropping py35 - so no rush.

blueyed:
> codecov/project — 51.54% (+3.71%) compared to 18b1a51

PR #994, opened 2020-02-02
show_call_signatures: doc, default to 2 without 'conceal'

blueyed:
It basically allows to give a warning when '1' is used / set explicitly, and
the 'conceal' feature is not available.

Also defaulting to '2' (cmdline mode) then seems better than not using
it.

This is just some minor clean up before
https://github.com/davidhalter/jedi-vim/pull/652.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -182,8 +182,10 @@ and usually saves one keypress.
 
 Jedi displays function call signatures in insert mode in real-time, highlighting
 the current argument. The call signatures can be displayed as a pop-up in the
-buffer (set to 1, the default), which has the advantage of being easier to refer
-to, or in Vim's command line aligned with the function call (set to 2), which
+buffer (set to 1 by default (with the conceal feature), 2 otherwise),
+which has the advantage of being easier to refer to (but is a hack with
+many drawbacks since it changes the buffer's contents),
+or in Vim's command line aligned with the function call (set to 2), which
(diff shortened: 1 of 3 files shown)


blueyed:
Somehow I would even recommend to use '2' by default, until https://github.com/davidhalter/jedi-vim/pull/652 is ready/merged already maybe?
(I am often running into issues with it when not using #652 myself)

davidhalter:
When does it not have conceal? Isn't that just extremely old VIM versions?

> Somehow I would even recommend to use '2' by default, until #652 is ready/merged already maybe?

IMO it works ok. I use `1` all the time. There are some issues, but I haven't really figured out what they are and I'm not sure #652 solves them.

blueyed:
> When does it not have conceal? Isn't that just extremely old VIM versions?

Likely.  But good anyway - and something similar is needed with #652 anyway.

> There are some issues, but I haven't really figured out what they are and I'm not sure #652 solves them.

One of the more annoying parts is when a word is (partly) covered through them and then `Ctrl-P` does not find it anymore for example.
Anyway, let's not change the default then.

davidhalter:
I don't really care about this change, because I don't think it really changes anything for users :) So feel free to merge yourself if you want to. I'm not against the patch.

blueyed:
Yeah, not functional changes are really expected.
But it also improves the docs a bit.

PR #996, opened 2020-02-25
Fix jedi#add_goto_window: use :copen to go to the qf window always

blueyed:
Without this it would not go there with an already open qf window, and
would go to an unexpected buffer line instead of selecting the entry.

This is a follow-up to b689409, where it was decided that it should
always require the user to select an entry.

An exception is when the same usages are used again: then it will select
the nearest/current entry only (via ":cc").


davidhalter:
Thanks a lot for working on this! I'm merging, because it's definitely the right direction. Will also test a bit soon.

PR #997, opened 2020-02-25
show_goto_multi_results: handle `full_name=None`

blueyed:
Seen with `<Definition name='key', description='param key'>` (using
usages on an function argument) [1].

1: https://github.com/pytest-dev/py/blob/2b6bd2925/py/_vendored_packages/iniconfig.py#L32-L33


davidhalter:
Makes a lot of sense. Thanks!

Issue #998, opened 2020-02-25
Vim freezing when opening python file


wviana:
### Issue

I'm currently unable to use jedi, cause sometime to now every time I do open a python file with it enabled it got stuck.

Currently I've tried to open vim with `vim -V --noplugins` and than source the `~/.vim/plugged/jedi-vim/autoload/jedi-vim.py` 

Got this output 

```
chdir(/Users/wviana/.vim/plugged/jedi-vim/autoload)
fchdir() to previous dir
sourcing "/Users/wviana/.vim/plugged/jedi-vim/autoload/jedi.vim"
Error detected while processing function jedi#init_python[4]..<SNR>17_init_python[27]..jedi#setup_python_imports:
line   25:
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "<string>", line 8, in <module>
ImportError: No module named jedi_vim_debug
finished sourcing /Users/wviana/.vim/plugged/jedi-vim/autoload/jedi.vim
Press ENTER or type command to continue
```

wviana:
After more test and reasearch I'm pretty sure it's  because my system have just python3.8, and looks like macvim does not support it. I'm thinking about keeping two versions, but really missing GNU/Linux right now.

davidhalter:
Hmmm from time to time we get weird error messages like this one for Mac users. Sometimes it's just a broken VIM installation. Some people just install a new VIM version from brew or whatever. Did you try that?

blueyed:
@wviana 
Try running the Python code per line manually: https://github.com/blueyed/jedi-vim/blob/76633a57837021c2d813dc1d23d889a88058844f/autoload/jedi.vim#L163-L172

Probably `import jedi_vim` fails already (and we fail then because we also try to import `jedi_vim_debug` then.
Also try/see if it works with `:py3` vs `:py` etc.

blueyed:
OTOH it looks like this for me, i.e. it includes the first error (py3 though only):
```
Error detected while processing function jedi#init_python[4]..<SNR>131_init_python[27]..jedi#
setup_python_imports:
line   26:
Traceback (most recent call last):
  File "<string>", line 3, in <module>
ModuleNotFoundError: No module named 'jedi_vimmm'
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "<string>", line 8, in <module>
ModuleNotFoundError: No module named 'jedi_vim_debuggg'
```

[...]

PR #1000, opened 2020-02-28
Improve/harden jedi#setup_python_imports

blueyed:
In particular it should not try to import `jedi_vim_debug` during error
handling "unprotected", which might hide the real error on Python 2 at
least.

Ref: https://github.com/davidhalter/jedi-vim/issues/998#issuecomment-592305240


davidhalter:
Makes sense, thanks! :)

Issue #1017, opened 2020-06-09
Feature request: specify buffer type per command


ethan-leba:
I have a small feature request that I'd like to take a stab at if it's a change you'd be willing to merge.

I use the splits mode for buffers in this plugin, which works great for jumping to definitions. However refactoring in this mode on a larger project will often end up creating 4+ splits which is an annoyance to deal with. So I propose these new configuration settings:

```
jedi#use_tabs_not_buffers_goto
jedi#use_splits_not_buffers_goto
jedi#use_tabs_not_buffers_refactor
jedi#use_splits_not_buffers_refactor
```
The above settings would override the current `g:jedi#use_(tabs/splits)_not_buffers` options if set, but otherwise the current option in place would not be modified.

Thanks for your work on jedi and jedi-vim!

davidhalter:
Is your problem just with refactoring?

ethan-leba:
Yes, that's what motivated me to file this issue.

davidhalter:
I think we should probably just not split in case a refactoring happens. IMO that's the best way forward. People that want splits want one split and not a lot of them.

davidhalter:
@blueyed What do you think?

ethan-leba:
Can I move forward with implementing this in the manner you suggested above @davidhalter?

blueyed:
@ethan-leba yeah, give it a try - I agree that we should probably not have different settings for this.

It might help to clarify your use case before, so we can try it ourselves based on your example/steps.
Do I understand it correctly that you are using `jedi#rename()` (`<leader>r`) with  `g:jedi#use_splits_not_buffers` set to e.g. "right"?
I've just tried it resulting in 9 splits.
I guess it depends on how they are auto-resized/displayed then.  I guess you are using `set equalalways` then?

I am not using this feature myself in general, but would like for it to display changed locations in a new quickfix list window instead of creating new windows in general (by default).
This would basically be what `show usages` does, but with the additional renaming.

(It would also be nice if there was a separate/new command to undo the changes in one step - as an idea, not necessary to improve it already / fix this issue, of course)

blueyed:
btw: I've noticed that the doc/logic is a bit off here with regard to the `g:jedi#use_splits_not_buffers` option:
The help says the default is `""`, but when you set that you get a warning about an unknown option.

[...]

PR #1020, opened 2020-06-27
Only use buffer on rename and show usages

ethan-leba:
Here's my attempt on #1017. It's mostly working, the only issue I ran into is that the error `E21: Cannot make changes, 'modifiable' is off` always pops up before the quickfix list opens to show usages after refactoring. Other than that, it works, but I'm pretty stumped on this bug so I thought I'd just open the PR and see if any of the maintainers know what's going on here!

### DIFF
--- a/pythonx/jedi_vim.py
+++ b/pythonx/jedi_vim.py
@@ -444,6 +444,7 @@ def show_goto_multi_results(definitions, mode):
             # well.
             lst.append(dict(text=PythonToVimStr(d.description)))
         else:
+            # EL: Error here
             text = annotate_description(d)
             lst.append(dict(filename=PythonToVimStr(relpath(d.module_path)),
                             lnum=d.line, col=d.column + 1,
@@ -1001,6 +1002,8 @@ def do_rename(replace, orig=None):
     # Save original window / tab.
     saved_tab = int(vim_eval('tabpagenr()'))
     saved_win = int(vim_eval('winnr()'))
+    saved_buffer = vim.current.buffer
+    saved_cursor = vim.current.window.cursor
 
     temp_rename = usages(visuals=False)
     # Sort the whole thing reverse (positions at the end of the line
@@ -1014,7 +1017,7 @@ def do_rename(replace, orig=None):
 
         if os.path.abspath(vim.current.buffer.name) != r.module_path:
             assert r.module_path is not None
-            result = new_buffer(r.module_path)
+            result = new_buffer(r.module_path, force_buffer=True)
             if not result:
                 echo_highlight('Failed to create buffer window for %s!' % (
                     r.module_path))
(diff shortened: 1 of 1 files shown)


ethan-leba:
Looks like master is failing CI

davidhalter:
Master passed: https://travis-ci.org/github/davidhalter/jedi-vim/builds/659627955. Not sure if something on the CI changed.

However, I'm still pretty sure that this PR doesn't have anything to do with the failure, so we should probably just change this test (it has been annoying before).

ethan-leba:
Sorry, worded that poorly. What I meant by that was that running tests on master locally failed for me

pythonx/jedi_vim.py
```diff
@@ -1027,11 +1030,15 @@ def do_rename(replace, orig=None):
...
 
+    vim_command('w')
+    vim.current.buffer = saved_buffer
+    vim.current.window.cursor = saved_cursor
     # Restore previous tab and window.
     vim_command('tabnext {0:d}'.format(saved_tab))
     vim_command('{0:d}wincmd w'.format(saved_win))
 
     if len(buffers) > 1:
+        usages()
```
davidhalter, line 1041: Why is this necessary?

pythonx/jedi_vim.py
```diff
@@ -1074,11 +1081,11 @@ def py_import_completions():
 
 
 @catch_and_print_exceptions
-def new_buffer(path, options='', using_tagstack=False):
+def new_buffer(path, options='', using_tagstack=False, force_buffer=False):
     # options are what you can to edit the edit options
-    if int(vim_eval('g:jedi#use_tabs_not_buffers')) == 1:
+    if int(vim_eval('g:jedi#use_tabs_not_buffers')) == 1 and not force_buffer:
```
davidhalter, line 1086: For tabs the old behavior is fine IMO. I use it regularly.

pythonx/jedi_vim.py
```diff
@@ -1027,11 +1030,15 @@ def do_rename(replace, orig=None):
         vim.current.buffer[r.line - 1] = (r_line[:r.column] + replace +
                                           r_line[r.column + len(orig):])
 
+    vim_command('w')
```
davidhalter, line 1033: I don't think we should do this. We should not save randomly. This is the user's responsibility.

pythonx/jedi_vim.py
```diff
@@ -444,6 +444,7 @@ def show_goto_multi_results(definitions, mode):
             # well.
             lst.append(dict(text=PythonToVimStr(d.description)))
         else:
+            # EL: Error here
```
davidhalter, line 447: What does this mean?

davidhalter:
CI should be working smooth now.

pythonx/jedi_vim.py
ethan-leba: You're right, will fix this :+1:

pythonx/jedi_vim.py
ethan-leba: Was following blueyed's suggestion here: https://github.com/davidhalter/jedi-vim/issues/1017#issuecomment-647227001

pythonx/jedi_vim.py
ethan-leba: That makes sense, but for creating new buffers the function `new_buffer` saves anyways for all buffers except the final one, and the buffer needs to be saved in order to return to the original buffer where the refactor happened.

ethan-leba: I should disable this behavior for tabs though

pythonx/jedi_vim.py
ethan-leba: That was where I thought the bug I was referring to in the description above was coming from, but I narrowed it down to this: https://github.com/davidhalter/jedi-vim/blob/master/autoload/jedi.vim#L484

ethan-leba: Still have no clue why an error is being emitted 😬

pythonx/jedi_vim.py
blueyed: Saving should at least get skipped with `set hidden` then maybe (a sane default).

PR #1035, opened 2020-12-07
Make use native popup windows

mymedia2:
This pull request offers to switch to native popup windows fully available since Vim 8.2. This solution will fix overlapping with other plugins which rely on buffer content. For example, [vim-gitgutter][1] is triggered by Jedi and marks lines above current one as modified due to invisible `?!?` stuff.

At the moment, the new implementation is not completed. It does not highlight selected parameter yet. If the idea makes interested, I will add full support to re-implement existing behavior.

 [1]: https://github.com/airblade/vim-gitgutter

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ $(dir $(BUILD_VIRTUAL_ENV)):
 	mkdir -p $@
 
 $(BUILD_VIRTUAL_ENV): | $(dir $(BUILD_VIRTUAL_ENV))
-	python -m venv $@
+	python3 -m venv $@
 
 $(BUILD_VIRTUAL_ENV)/bin/vint: | $(BUILD_VIRTUAL_ENV)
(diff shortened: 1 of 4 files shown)


davidhalter:
I'm definitely interested in that kind of stuff. Does that also work in neovim?

If the `?!?` is annoying for you I would recommend using `show_call_signatures = 2` or something like this.

mymedia2:
I fear NeoVim does not support this API, but it seems to have [api-floatwin][1] that may be used for comparable functionality.

 [1]: https://neovim.io/doc/user/api.html#api-floatwin

blueyed:
This overlaps with https://github.com/davidhalter/jedi-vim/pull/652.

blueyed:
@mymedia2 please consider trying https://github.com/davidhalter/jedi-vim/pull/652

PR #1042, opened 2021-02-03
Add install instruction for plain Vim 8

abitrolly:


davidhalter:
I would rather not have all those different ways of installing jedi-vim. I would probably just recommend https://github.com/junegunn/vim-plug and remove the other options. @blueyed can decide.

blueyed:
Yeah, having all the different options of installment are a recurring issue with READMEs across (Neo)Vim plugins.
I think people in general should be aware how to install a plugin based on some GitHub URL/slug.

Here however it is a nuisance that submodules are used, which are not handled with every package manager, or when using no package manager at all.

I agree in general that the README section should be revisited probably, and that it should contain instructions for Vim and NeoVim, which both have support for packages (which this change is about apparently), and then maybe vim-plug as an example for a plugin manager.

README.rst
```diff
@@ -79,12 +79,17 @@ You need a VIM version that was compiled with Python 2.7 or later
 
 Manual installation
 -------------------
+The plugin uses `git` submodules to fetch latest version of Jedi.
```
blueyed, line 82:
```suggestion
The plugin uses `git` submodules to fetch the vendored versions of Jedi.
```
(I've figured it would be better to not include details about "and its dependencies (parso)")

README.rst
```diff
@@ -79,12 +79,17 @@ You need a VIM version that was compiled with Python 2.7 or later
...
-You might want to use `pathogen <https://github.com/tpope/vim-pathogen>`_ or
-`Vundle <https://github.com/gmarik/vundle>`_ to install jedi-vim.
+To install using standard Vim 8 plugin system:
+
+.. code-block:: sh
 
-The first thing you need after that is an up-to-date version of Jedi. Install
-``git submodule update --init --recursive`` in your jedi-vim repository.
+    mkdir -p ~/.vim/pack/git-plugins/start
+    git clone --recursive https://github.com/davidhalter/jedi-vim.git ~/.vim/pack/git-plugins/start/jedi-vim
```
blueyed, line 89: btw: do you know if there is a preferred/recommended way to update Vim packages?

README.rst
```diff
@@ -79,12 +79,17 @@ You need a VIM version that was compiled with Python 2.7 or later
...
+
+Or you might want to use `pathogen <https://github.com/tpope/vim-pathogen>`_ or
+`Vundle <https://github.com/gmarik/vundle>`_ to install jedi-vim.
```
blueyed, line 92: We can merge this with the rest to only include instructions for vim-plug then.

blueyed (commented):
Thanks for attacking this.

I think it should get revisited, and could do so myself based on the given feedback, but feel free to either do it yourself, or pick it up (@davidhalter).

README.rst
davidhalter: no :D

davidhalter:
@blueyed Feel free to merge whatever here.

README.rst
blueyed: @abitrolly do you know?  How do you keep them updated?

(as for me: I am using vim-plug basically, but given the amount of local changes I still update them manually in general, i.e. by updating/merging into the Git checkout)

README.rst
blueyed: @abitrolly what do you think?  Should I/we go ahead and do so here?

README.rst
abitrolly: I update with plain `git`.
```
git -C ~/.vim/pack/git-plugins/start/jedi-vim up
```

README.rst
abitrolly: I am not a user of `vim` plugin managers, so won't rely on my opinion towards them.

PR #1048, opened 2021-03-15
Support adding extra sys_path

SeungChan92:


davidhalter:
Thanks!

doc/jedi-vim.txt
```diff
@@ -518,6 +518,14 @@ Both setting the directory and setting a project is working.
...
 Default: "auto"
 
+------------------------------------------------------------------------------
+6.15. `g:jedi#added_sys_path`                *g:jedi#added_sys_path*
+
+To add extra sys_path.
+
+Examples: ["../site-packages"]
+Default: []
+
```
blueyed, line 528: Where is this relative to then?
I think it is worth explaining.

PR #1059, opened 2021-06-21
The buffer-local variable `b:jedi_environment_path`

jamescherti:
The buffer-local variable `b:jedi_environment_path` was added to jedi-vim to allow the user to override the global variable `g:jedi#environment_path` in the current buffer (useful to set a different virtualenv or Python version in the current buffer).

### DIFF
--- a/doc/jedi-vim.txt
+++ b/doc/jedi-vim.txt
@@ -45,6 +45,7 @@ Contents                                *jedi-vim-contents*
     6.13. smart_auto_mappings           |g:jedi#smart_auto_mappings|
     6.14. use_tag_stack                 |g:jedi#use_tag_stack|
     6.15. environment_path              |g:jedi#environment_path|
+                                        |b:jedi_environment_path|
     6.16. added_sys_path                |g:jedi#added_sys_path|
 7. Testing                              |jedi-vim-testing|
 8. Contributing                         |jedi-vim-contributing|
@@ -510,6 +511,7 @@ Default: 1 (enabled by default)
 
 ------------------------------------------------------------------------------
 6.15. `g:jedi#environment_path`                *g:jedi#environment_path*
+                                               *b:jedi_environment_path*
 
 To use a specific virtualenv or a specific Python version it is possible to
 set an interpreter.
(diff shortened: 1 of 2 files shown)


davidhalter (approved):
IMO this looks good. Thanks @jamescherti 

Is there any other feedback? Otherwise I will probably merge soon.

bew:
Is it possible to name it `b:jedi#environment_path` for consistency with the global?

davidhalter:
@bew  Good catch! I personally didn't see that. I agree that would probably make sense. IMO it's still a mistake that the hash is in there, but it's in there everywhere. We should probably try to be consistent.

jamescherti:
@bew @davidhalter I have tried to name it `b:jedi#environment_path` to be consistent with global variables, but I received the following error message from Vim:
```
:let b:jedi#environment_path="/home/user/.venv"
E416: Illegal variable name: b:jedi#environment_path
```

jamescherti:
@davidhalter @bew According to the Vim documentation https://vimhelp.org/eval.txt.html#internal-variables : "An internal variable name can be made up of letters, digits and '_'.  But it cannot start with a digit.". Internal variables include buffer and global variables.  

The documentation does not seem mention that characters like hashes can be used in variable names. It may explain the Vim error message: "E461: Illegal variable name: b:jedi#environment_path".

I've also noticed that the developers of Vim plugins (like Ale) that allow users to use buffer-local variable to override global variables, chose to use underscores instead of hashes in their naming convention. I wonder if it is because they had the same issue?

**Here are the solutions I suggest:**
Solution 1 - Replace "g:jedi#" variables with "g:jedi_" (and keep backward compatibility with "g:jedi#").
Solution 2 - Use a different naming convention for buffer-local variables: "b:jedi_" (not consistent with global variable, but easier to implement).

What do you think?

bew:
I'd say solution 2 for simplicity and to keep this pr on point.
If @davidhalter wants to change all option names (with fallback) to use underscores it can be done in a later pr.

davidhalter:
@jamescherti I agree with @bew, leave it as it is for now. I did not know that it was not possible, sorry for that.

-----

I wanted to merge, but there was an issue with the tests, I think it's related. Right?

jamescherti:
@davidhalter @bew The following exception was fixed in the last commit:
```
jedi_vim.VimError: Vim(let):E121: Undefined variable: b:jedi_environment_path; created by 'b:jedi_environment_path'
```

You should be able to merge now.

davidhalter:
Thanks a lot!

jamescherti:
Thank you @davidhalter and @bew for the quick approval and merge!

PR #1060, opened 2021-06-22
doc: Add hint about python 2 support

kozachynskyi:
I searched for half an hour, why in :checkhealth of nvim, python 2 virtual environment is not shown, until I realized, that py2 is not supported by jedi itself.

So I think it's a good idea to mention in Readme that python 2 is not supported anymore. Maybe saves someones time.


davidhalter:
Thanks!

PR #1068, opened 2021-11-17
Update jedi to v0.18.1 and parso to v0.8.2

jgb:
The update of jedi specifically fixes a bug which prevents completion from working as it should (https://github.com/davidhalter/jedi/issues/1817)
I update parso to the latest tag for good measure.


davidhalter:
Thanks! :)

PR #1072, opened 2022-01-16
Fix added_sys_path default setting

ThomasBlauth:
I think I found a small bug in the default settings. The default setting for `added_sys_path` is sting quoted twice which leads to the value of `g:jedi#added_sys_path` beeing the string `"[]"` instead of an empty list.

Executing `python3 print(jedi_vim.get_project().added_sys_path)` returns `['[', ']']` instead of an empty list.

A results, for example, is that `:checkhealth` crashes when `g:jedi#added_sys_path` is actually set.

```
:let g:jedi#added_sys_path = ["../test"]

:checkhealth

 health#jedi#check
========================================================================
  - ERROR: Failed to run healthcheck for "jedi" plugin. Exception:
    function health#check[21]..health#jedi#check[2]..jedi#debug_info, line 64
    Vim(if):E691: Can only compare List with List
```

I'm not too familiar with vimscript, I hope my assumtions are actually true and not a misunderstanding on my part!

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -55,5 +55,6 @@ Brian Mego (@brianmego)
 Stevan Milic (@stevanmilic) <stevan.milic@yahoo.com>
 Konstantin Glukhov (@Konstantin-Glukhov)
 Seungchan An (@SeungChan92) <dev.issea1015@gmail.com>
+Thomas Blauth (@ThomasBlauth) <thomas.blauth@protonmail.com>
 
 @something are github user names.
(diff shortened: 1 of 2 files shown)


davidhalter:
Yes, this makes sense. Thanks for finding it! There's an issue with the CI though (the string should probably be `'[]'` instead of `"[]"`).

davidhalter:
Thanks! (There's something wrong with the coverage CI, but that seems to have nothing to do with this PR.)

PR #1073, opened 2022-01-17
get_project: simplify getting b:jedi_environment_path override

blueyed:
- the common case is having an exception here (with no buffer setting),
  which results in the exception handler always
- vim_eval has overhead, especially with Neovim, so the less calls the
  better

This patch changes it to query the `b:` dict, using the (always
set/initialized) global setting as the default.

Ref: #1059 /cc @jamescherti (just for information)

### DIFF
--- a/pythonx/jedi_vim.py
+++ b/pythonx/jedi_vim.py
@@ -218,13 +218,9 @@ _current_project_cache = None, None
 
 
 def get_project():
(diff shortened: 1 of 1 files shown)


davidhalter:
Thanks! CI is failing, because of the coverage part, which is never completing.

PR #1076, opened 2022-01-29
Add option to enable/disable case insensitive completion

jamescherti:
This pull request adds an option to enable/disable case insensitive completion (the Jedi setting: `jedi.settings.case_insensitive_completion`).

It adds the following options to jedi-vim:
- Global option: `g:jedi#case_insensitive_completion`
- Buffer-local option: `b:jedi_case_insensitive_completion`

Values:
- 0 to disable case insensitive completion.
- 1 to enable case insensitive completion (default).


davidhalter:
Thanks!

blueyed:
@davidhalter 
Is there something like smart-case in Jedi, i.e. use case-sensitivity if there is any upper case character in the input, but otherwise case-insensitivity?
(this allows to enable case-sensitivity "on the fly")

davidhalter:
No there is currently no such thing. One could obviously implement that in Jedi (or here).

PR #1082, opened 2022-04-05
Add the buffer-local option 'b:jedi_added_sys_path'

jamescherti:
This pull request adds the buffer-local option `b:jedi_added_sys_path` that can be used to add additional sys_path.

### DIFF
--- a/pythonx/jedi_vim.py
+++ b/pythonx/jedi_vim.py
@@ -222,7 +222,12 @@ def get_project():
         "get(b:, 'jedi_environment_path', g:jedi#environment_path)"
     )
     vim_project_path = vim_eval("g:jedi#project_path")
-    vim_added_sys_path = vim_eval("g:jedi#added_sys_path")
+    vim_added_sys_path = vim_eval(
+        "get(g:, 'jedi#added_sys_path', [])"
+    )
+    vim_added_sys_path += vim_eval(
+        "get(b:, 'jedi_added_sys_path', [])"
+    )
 
     global _current_project_cache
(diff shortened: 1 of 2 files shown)


pythonx/jedi_vim.py
```diff
@@ -222,7 +222,12 @@ def get_project():
         "get(b:, 'jedi_environment_path', g:jedi#environment_path)"
     )
     vim_project_path = vim_eval("g:jedi#project_path")
-    vim_added_sys_path = vim_eval("g:jedi#added_sys_path")
+    vim_added_sys_path = vim_eval(
+        "get(g:, 'jedi#added_sys_path', [])"
+    )
+    vim_added_sys_path += vim_eval(
+        "get(b:, 'jedi_added_sys_path', [])"
+    )
```
davidhalter, line 230: There's a bug in the logic below that gets even worse when we start using buffer definitions.

The problem is that `vim_added_sys_path` is not part of the `cache_key`. Would you agree? I feel like we should add it there as well.

jamescherti: @davidhalter I agree, David. Please take a look at the latest commit. I have just added 'g:jedi#added_sys_path' and 'b:jedi_added_sys_path' to 'cache_key'.

davidhalter:
Great stuff, thanks! `load_project` should probably be refactored to use `get_project`, but that should not be your concern.

jamescherti:
> Great stuff, thanks! 

My pleasure!

> `load_project` should probably be refactored to use `get_project`, but that should not be your concern.

Good idea, David. Making `load_project()` use `get_project()` will reduce code duplicate.

Issue #1092, opened 2022-11-10
use of popup instead of preview window 


Konfekt:
Since Patch 8.1.1882 Vim offers the option `popup` superseding `preview` in `&completeopt` to show a popup instead of preview window on completion. 
However, with latest Jedi-Vim from Git Master

```vim
let g:jedi#auto_vim_configuration=0
set completeopt=menuone,longest,popup
```

won't show a popup window.
Instead,

```vim
let g:jedi#auto_vim_configuration=0
set completeopt=menuone,longest,popup,preview
```

does, even though `popup` supersedes `preview`.

This is likely due to

```python
#  pythonx/jedi_vim.py (lines 363-379)
add_info = "preview" in vim.eval("&completeopt").split(",")
out = []
for c in completions:
    d = dict(word=PythonToVimStr(c.name[:len(base)] + c.complete),
             abbr=PythonToVimStr(c.name_with_symbols),
             # stuff directly behind the completion
             menu=PythonToVimStr(c.description),
             icase=1,  # case insensitive
             dup=1  # allow duplicates (maybe later remove this)
             )
    if add_info:
        try:
            d["info"] = PythonToVimStr(c.docstring())
        except Exception:
            print("jedi-vim: error with docstring for %r: %s" % (
                c, traceback.format_exc()))
    out.append(d)
```

How about checking here for `popup` as well?

davidhalter:
I did not know that. 100% agreed. Can you PR this and add the default of `popup` as well in `plugin/jedi.vim`. This is much much better than the preview window.

PR #1093, opened 2022-11-10
prefer popup over preview window

Konfekt:
addresses #1092

### DIFF
--- a/ftplugin/python/jedi.vim
+++ b/ftplugin/python/jedi.vim
@@ -44,10 +44,15 @@ if g:jedi#auto_initialization
         inoremap <silent> <buffer> <space> <C-R>=jedi#smart_auto_mappings()<CR>
     end
 
-    if g:jedi#auto_close_doc
+    if g:jedi#auto_close_doc && (&g:completeopt =~# '\<preview\>' && &g:completeopt !~# '\<popup\>')
         " close preview if its still open after insert
         augroup jedi_preview
-            autocmd! InsertLeave <buffer> if pumvisible() == 0|pclose|endif
+            if v:version > 704
+                autocmd CompleteDone <buffer> * pclose
+            else
+                autocmd InsertLeave <buffer> * if pumvisible() == 0|pclose|endif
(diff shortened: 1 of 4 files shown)


ftplugin/python/jedi.vim
```diff
@@ -44,10 +44,15 @@ if g:jedi#auto_initialization
         inoremap <silent> <buffer> <space> <C-R>=jedi#smart_auto_mappings()<CR>
     end
 
-    if g:jedi#auto_close_doc
+    if g:jedi#auto_close_doc && (&g:completeopt =~# '\<preview\>' && &g:completeopt !~# '\<popup\>')
         " close preview if its still open after insert
         augroup jedi_preview
-            autocmd! InsertLeave <buffer> if pumvisible() == 0|pclose|endif
+            if v:version > 704
```
davidhalter, line 50: What's the reasoning for this change?

davidhalter:
I like the changes a lot, I'm just unsure why you changed that one specific thing.

ftplugin/python/jedi.vim
Konfekt: Hello, 

I just implemented [the first two answers](https://stackoverflow.com/questions/3105307/how-do-you-automatically-remove-the-preview-window-after-autocompletion-in-vim) as `:help g:jedi#auto_close_doc` states that

> this window is being closed after insertion of a completion item.

the `CompleteDone` event seemed the more exact autocmd to achieve this, whenever available.

davidhalter: Thanks that makes sense.

davidhalter:
The coverage tests are currently failing, I feel like this is something unrelated. Maybe @blueyed has any idea? I feel like if he does not chime in I might disable the coverage tests temporarily.

davidhalter:
Sorry forgot to merge, because I spilled tea on my notebook and it took me a while to get it back in good shape. Now that I have it back, I found the tab of this PR again. Thanks a lot again :)

Issue #1094, opened 2022-11-22
Error: "E16: Invalid range: * pclose"


zakutin:
### Issue

After updating the jedi-vim plugin with 28a6765, df261c0, 40c7f6d, 8c7161f, the error message appears on autocompletion.
```
Error detected while processing CompleteDone Autocommands for "<buffer=1>":
E16: Invalid range: * pclose
```
### Steps to reproduce
1. Update the jedi-vim with 28a6765, df261c0, 40c7f6d, 8c7161f (I have no idea which commit exactly causes the error);
2. Open python file;
3. Type any known function, press ctrl+space, select required function, press space/enter/arrow right;
4. Error appears;
5. Revert the update in vim-plug — autocompletion works without any errors.

### Output of “:verbose JediDebugInfo”
<details>

##### jedi-vim version

 - jedi-vim git version: 0.11.1
 - jedi git submodule status:  fae26fa7a427b94a21e11500503d8008647e6309 pythonx/jedi (v0.18.1)
 - parso git submodule status:  ee5edaf22ff3941cbdfa4efd8cb3e8f69779fd56 pythonx/parso (v0.8.3)

##### Global Python

Using Python version 3 to access Jedi.
 - global sys.executable: `/usr/bin/python3`
 - global sys.version: `3.10.6 (main, Nov  2 2022, 18:53:38) [GCC 11.3.0]`
 - global site module: `/usr/lib/python3.10/site.py`

##### Jedi

 - path: `/home/user/.vim/plugged/jedi-vim/pythonx/jedi/jedi/__init__.py`
 - version: 0.18.1

##### Jedi environment: <SameEnvironment: 3.10.6 in /usr>

 - executable: /usr/bin/python3
 - sys_path:
    - `/usr/lib/python310.zip`
    - `/usr/lib/python3.10`
    - `/usr/lib/python3.10/lib-dynload`
    - `/home/user/.local/lib/python3.10/site-packages`
    - `/usr/local/lib/python3.10/dist-packages`
    - `/usr/lib/python3/dist-packages`

##### Known environments

 - <Environment: 3.10.6 in /usr> (/usr/bin/python3.10)

##### Settings

```
g:jedi#popup_on_dot = 0 (default: 1)
g:jedi#use_tabs_not_buffers = 1 (default: 0)
g:jedi#show_call_signatures = '2' (default: 1)


  omnifunc=jedi#completions
[...]

PR #1095, opened 2022-11-23
fix typo

Konfekt:
Fix https://github.com/davidhalter/jedi-vim/issues/1094


davidhalter:
Thanks, that helps a lot!

PR #1097, opened 2022-12-23
Add a keyboard mapping to rename  variables/functions/classes without deleting the word under the cursor

jamescherti:
This pull request adds the key mapping `<Leader>R` (upper case R) to rename variables/functions/classes without deleting the word under the cursor:
- **`<Leader>R` in Normal mode:** Jedi-vim keeps the word under the cursor, moves the cursor to the end of the word, and puts Vim in insert mode, where the user is expected to enter the new variable name.
- **`<Leader>R` in Visual mode:** Use the selected text as the default answer to the prompt.

**Bug fix:** This pull request also adds the support for `set selection=exclusive` to the function `rename_visual()`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -207,6 +207,7 @@ get more information. If you set them to ``""``, they are not assigned.
     let g:jedi#usages_command = "<leader>n"
     let g:jedi#completions_command = "<C-Space>"
     let g:jedi#rename_command = "<leader>r"
+    let g:jedi#rename_command_keep_name = "<leader>R"
 
 An example for setting up your project:
 
--- a/autoload/jedi.vim
+++ b/autoload/jedi.vim
@@ -24,6 +24,7 @@ let s:default_settings = {
     \ 'call_signatures_command': "'<leader>n'",
     \ 'usages_command': "'<leader>n'",
     \ 'rename_command': "'<leader>r'",
+    \ 'rename_command_keep_name': "'<leader>R'",
     \ 'completions_enabled': 1,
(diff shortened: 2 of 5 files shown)


davidhalter:
So that would just not delete the word under the cursor initially, but otherwise work pretty much the same? Because we have been discussing internally to have two different ways of doing renames anyway.

jamescherti:
That is right, @davidhalter . The behavior is the same as before, but if you set `g:jedi#rename_deletes_word` or `b:jedi_rename_deletes_word` to 1, the rename command will not delete the word under the cursor anymore. It will move the cursor to the end of the word and switch to insert mode, allowing the user to edit it.

This option will make it easier for the user to edit long variables.

jamescherti:
**FYI:** I added the support for `set selection=exclusive` to the function `rename_visual()` (bug fix) and changed the option name to `g:jedi#rename_command_reuse_name` and `b:jedi_rename_command_reuse_name`.

davidhalter:
Hmmm I feel like we should probably have a different command for what you are trying to accomplish. Currently the default is `<leader>r` for rename and I feel like we should do something like a `<leader>R` to do what you want. Because I agree with what you are trying to achieve, I would just prefer a different command entirely.

jamescherti:
Good idea, @davidhalter . Please check the latest commit.

davidhalter:
Tested and looks good to me, thanks a lot!

jamescherti:
Thank you for merging!

PR #1098, opened 2022-12-26
Open the pydoc documentation in a new tab

jamescherti:
This pull request adds a feature that allows opening the pydoc documentation in a new tab when the user presses the `K` key.

It also adds a new option to choose how Jedi-vim opens the documentation:

_**Option:**_ `g:jedi#way_to_open_documentation` (The buffer-local version: `b:jedi_way_to_open_documentation`)

_**Option values:**_
- **split** (Open the documentation in a horizontal split)
- **tab** (Open the documentation in a new tab)

_**Default value:**_ split _(Same as before)_

### DIFF
--- a/autoload/jedi.vim
+++ b/autoload/jedi.vim
@@ -27,6 +27,7 @@ let s:default_settings = {
     \ 'completions_enabled': 1,
     \ 'popup_on_dot': 'g:jedi#completions_enabled',
     \ 'documentation_command': "'K'",
+    \ 'way_to_open_documentation': "'split'",
     \ 'show_call_signatures': has('conceal') ? 1 : 2,
     \ 'show_call_signatures_delay': 500,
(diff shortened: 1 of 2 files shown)


davidhalter:
I do not think this is currently a good idea to add. The maintenance burden on stuff like this is just too big and I doubt a lot of people will use it (Most people do not read the docs anyway). But since jedi-vim isn't changing a lot, I feel like you can safely maintain a fork and just merge all newer changes.

Still thanks a lot for contributing.

jamescherti:
Thank you for taking the time to review the code, @davidhalter .

I have created the Vim plugin **[vim-jedipydoc](https://github.com/jamescherti/vim-jedipydoc)** that allows opening the Pydoc/Python documentation in a **new tab** (By default) or a **new horizontal split**.

davidhalter:
That sounds like a great solution, thanks!

PR #1106, opened 2023-04-11
update jedi to v0.18.2

jgb:
Hi,

as I use jedi-vim quite heavily I would like to update the git submodule of jedi to the newest available version.

Greetings,

jgb


davidhalter:
Thanks a lot!

PR #1117, opened 2023-12-27
Replace deprecated checkhealth syntax with newer syntax

dundargoc:


davidhalter:
Can we do this without lua? My current VIM for example doesn't have lua support.

dundargoc:
`:checkhealth` doesn't even work on vim since it's a neovim-exclusive command, and neovim always ships with lua.

davidhalter:
Sorry, this is not code I have written. I'm trying my best to understand.

So do I understand this correctly: This whole function is only about nvim and will never be invoked by VIM? I don't understand the mechanism.

dundargoc:
> This whole function is only about nvim and will never be invoked by VIM?

Yeah. Try running `:checkhealth jedi` in vim. Nothing will happen. This is a neovim-only functionality so it should be fine to use lua functionality through vimscript as lua is shipped with neovim.

davidhalter:
Thanks!

PR #1124, opened 2024-07-06
Add Neovim Installation Requirements

adamency:
Add information for Neovim that the python package `pynvim` is needed for a working python environment inside `nvim`.

Also so provide command needed to check that the environment setup is functional.


davidhalter:
Thanks!

PR #1130, opened 2024-12-27
update jedi to v0.19.2 and parso to v0.8.4

jgb:


davidhalter:
Thanks. Unfortunately somebody needs to look at the tests CI. Somethin g seems wrong, but it might as well be test. I would have to investigate.

BTW: Ignore the coverage CI, it should probably just be removed, it has been defunctional for a long time.

jgb:
> Thanks. Unfortunately somebody needs to look at the tests CI. Somethin g seems wrong, but it might as well be test. I would have to investigate.
> 
> BTW: Ignore the coverage CI, it should probably just be removed, it has been defunctional for a long time.

I'm running the tests locally, it seems it's the jedi upgrade that causes the tests to fail, upgrading parso alone doesn't cause any harm.
I'll try to see if it's a real problem or just an adjustment of the test is needed.

davidhalter:
I'll check. Thanks for looking!

The tests are sometimes a bit brittle here and react to non-problems, so don't worry.

davidhalter:
I fixed a few things around CI and basically just removed test parts that did not make sense: https://github.com/davidhalter/jedi-vim/pull/1131

Thanks for your PR again.

PR #1133, opened 2025-01-12
Make minor effeciency changes to regex usages

sirosen:
- For static patterns which are used repeatedly, compile them in module scope
- For identifier char checking, simply use a string collection membership check

These changes make jedi-vim faster by a small degree every time completions, rename, and show_call_signatures are invoked.

---

Ideally I'd have been able to give benchmarks to prove it's faster, but I haven't been able to get vspec working locally and therefore haven't gotten far on a local testing setup.
If these changes are unwanted, please don't hesitate to tell me. I'm happy to focus my attention elsewhere (though my vimscript is pretty weak).

### DIFF
--- a/pythonx/jedi_vim.py
+++ b/pythonx/jedi_vim.py
@@ -10,6 +10,7 @@ import re
 import os
 import sys
 from shlex import split as shsplit
+import string
 from contextlib import contextmanager
 from pathlib import Path
 try:
@@ -26,6 +27,10 @@ if is_py3:
 else:
     ELLIPSIS = u"…"
 
+IDENTIFIER_CHARS = set(string.letters + string.digits + "_")
+RIGHTMOST_WORD_PATTERN = re.compile(r"\w+$")
+SIGNATURE_QUOTES_PATTERN = re.compile(r'''\\*["']+''')
+RIGHTMOST_SIGNATURE_QUOTES_PATTERN = re.compile(r'''\\*["']+$''')
 
 try:
     # Somehow sys.prefix is set in combination with VIM and virtualenvs.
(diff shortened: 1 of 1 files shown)


davidhalter:
I'm pretty sure that `string.letters + string.digits + "_"` does not match the behavior of `\w` (besides the fact that the tests are failing).

I guess I am going to close for two reasons:

1. This won't have a significant impact on performance, since Jedi itself will always be slower by factors of magnitude.
2. While the change itself is fine in general, jedi-vim is unfortunately not well tested at all. So with these changes there is sometimes the risk that we overlook something that might cause bugs. Since jedi-vim is in maintenance mode and not used a lot anymore (most people use LSP these days for good reasons, even when using Jedi), it's IMO just an unnecessary risk.
3. I have limited time. Testing this and other changes well is usually not worth it if it doesn't have a significant upside.

So I still want to thank you for trying, I don't think that this is a bad change at all and we probably should have done it like that in the beginning, but it's just not worth the hassle.

sirosen:
100% understood! I think that's a more than fair response.

I was trying to get my hands dirty with some valid improvement before trying to help in other ways, not put an extra burden on you. I'll keep an eye out for ways I can help where "the juice is worth the squeeze", so to speak.

Thanks (as always) for the awesome tools!

davidhalter:
`\w` is documented like

```
For Unicode (str) patterns:

    Matches Unicode word characters; this includes all Unicode alphanumeric characters (as defined by str.isalnum()), as well as the underscore (_).

    Matches [a-zA-Z0-9_] if the ASCII flag is used.
```

Which in this case includes a big amount of unicode letters.

sirosen:
You're 100% right. I was being clumsy in my thinking and assumed that identifiers are ascii-only. But they are not.

PR #1138, opened 2025-08-24
Update parso to v0.8.5

jgb:


davidhalter:
I'm a bit confused why the CI does not run... But I guess I'll wait.

davidhalter:
I closed and reopened and it still won't run. Maybe some defaults are wrong?

davidhalter:
Can you rebase? I think CI should run now.

It was because the runner was still configured as ubuntu-20.04. I'm very surprised that this did not show a message like "No runner found for ubuntu-20.04"....

PR #1146, opened 2026-06-16
Saving initial column offset for completions in case a chain of

pablogipi:
The new **o** flag for complete allows to query completions from omni func after the next or previous word completion:
This is not working at the moment in jedi-vim because in completions() the tool first ask for current row and column to check
if it needs to find start, then calls it again, but if it is a chained completion lookup, like when adding the o flag this second callback to get current row and columns returns a wrong value for column because the first completion looks like is moving the cursor.
So here I'm saving the original column values in a buffer variable for later use and then all works as expected, the omnifunc works and adds it's results to the next/prev word search.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -57,5 +57,6 @@ Konstantin Glukhov (@Konstantin-Glukhov)
 Seungchan An (@SeungChan92) <dev.issea1015@gmail.com>
 Thomas Blauth (@ThomasBlauth) <thomas.blauth@protonmail.com>
 James Cherti (@jamescherti)
+Pablo Gimenez Pizarro (@pablogipi) <pablogipi@gmail.com>
 
 @something are github user names.
--- a/pythonx/jedi_vim.py
+++ b/pythonx/jedi_vim.py
@@ -342,9 +342,12 @@ def completions():
             if not re.match(r'[\w\d]', char):
                 break
             count += 1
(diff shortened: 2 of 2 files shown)


davidhalter:
Thanks for trying to contribute! Is there a reason you're not using LSP/Zuban? I think it's generally better at this point.

About the pull request: I have to find time to test and understand this. I'm not sure what it exactly fixes and I currently don't really have time for jedi-vim.

pablogipi:
Hey, sorry for the late response.
No, there  is no more reason to move into something like LSP/Zuban apart than just used to my setup for years now :)
I will move to LSP eventually.
What this fixes is how the completion function for _omnicomplete_ works in jedi. In vim 9.1 (don't remember the path number) a new option was added:
https://github.com/vim/vim/issues/9005
https://github.com/vim/vim/pull/17065

These new options allows to compose the complete results from Ctrl-N and omnifunc in one call to complete().
The problem is that the jedi completion code look at the current column and row twice, the complete func is called twice in vim, first to get the start of the completion word and second to return the potential results. This works fine when one call to omnifunc is done, the problem is that using the new **_o_** flag to compose next keywords complete and omnifunc causes the second call to get current row and column to return a wrong value, the cursor has been moved already by the first complete call for next keyword, and is also redundant since we just need the row and column results from the first callback, when the complete mechanism just tried to get the current word under the cursor.
This PR fixes that and allows jedi completion to work with these new flags in vim complete options.

